PR #2, opened 2016-08-16
Added dummy static src

bartaz:
Added dummy static index.html with normalize.css and some base styles (copied from SCA).

The stricture will likely change a bit when webpack kicks in, but it gives us the base to work on and discuss the styles.


bartaz:
The biggest questions I see when working on this PR (and assuming we are going to use CSS modules):

* do we use normalize.css at all?
* do we have any 'global' CSS (such as base.css) or is everything in modules?
* do we have any 'static' parts of template (such as header, footer) or is everything a React component?
* do we still use BEM and `b-` prefix or CSS modules do all the magic we need?

katiefenn:
- I think we should use normalize.css. The total filesize is small when gzipped, and I expect it will save us some declarations in module stylesheets. Also, it might come in handy if we're supporting older versions of Internet Explorer.
- If we're building the entire page with React, I think our stylesheets belong in modules.
- I think everything will be a React component.
- CSS Modules will namespace classnames for us, so we won't have to do this manually. But let it inspire your class naming if you're stuck.

earnubs:
The only thing I'll add is that CSS modules make the namespace "area" smaller, but you can still end up in a muddle on a complex component, so from my limited experience keeping to a naming pattern like BEM is good, prefixing (b-) isn't so useful here I don't think ... prefixing is super useful in something where you might need to search a bunch of html, js, python files for classnames you want to change ... but here our classnames in the css are decoupled from the class name in the js source or output, correct?

earnubs:
I've pushed a PR #3 with the HTML entirely generated from React components, I have done a couple of apps in my hobby time were the base templates were Nunjucks or hbs, but never one with everything in React. I will say that it can be very confusing having some stuff in a nunjucks template and most other stuff in React, so I would vote for everything in React at this time.

bartaz:
Yes, I agree that `b-` prefix doesn't make sense in this project, as in SCA it was mostly used to differentiate old CSS from new BEM classes.

As PR #3 is much closer to what we need from environment this one is redundant in terms of code.
I will add normalize css and header component into webpack based project and close this PR.

PR #3, opened 2016-08-16
basic server with webpack, hot module replacement, and tests

earnubs:
To start the server:

`npm start`

To run the test(s):

`npm test`

to watch files for changes and run tests:

`npm test:watch`


server/server.js
```diff
@@ -0,0 +1,42 @@
...
+import webpackHotMiddleware from 'webpack-hot-middleware';
+import { renderToString } from 'react-dom/server';
+
+import Html from '../src/helpers/Html';
+import webpackConfig from '../webpack/dev-config';
+
+const app = Express();
+const compiler = webpack(webpackConfig);
+
+if (!(process.env.NODE_ENV === 'production')) {
```
katiefenn: In the past we specifically tested for 'development' & 'staging' environments so that apps default to production settings if something unexpected happens to system configuration.

bartaz:
@earnubs I guess we may want to .gitignore public/static contents?

server/server.js
earnubs: Yep, this line shouldn't be here since what I'm proposing is the very bare bones of a dev server with no production bits, that can all come in follow on branches.

earnubs:
@bartaz i couldn't make my mind up if we want to ignore bundles or not, but for this development only branch we do ... for a prod setup it's possible you might want to be able to start the server and have things work without running webpack but the more i think about it, webpack would run on a postinstall hook from npm, sooo ... yeah i'll ignore it ;)

server/server.js
```diff
@@ -0,0 +1,39 @@
+import Express from 'express';
+import React from 'react';
+import ReactDOM from 'react-dom/server';
+import util from 'util';
+import webpack from 'webpack';
+import webpackDevMiddleware from 'webpack-dev-middleware';
+import webpackHotMiddleware from 'webpack-hot-middleware';
+import Html from '../src/helpers/Html';
```
bartaz: I found it a little bit confusing that html.js is in src/ while it is used in server/.

I know that distinction between server-side and front-end code is quite vague in project like this, and will get even more blurry when we get to full server-side rendering, so maybe it's not such an issue.

server/server.js
earnubs: I've now dropped this line, prod v dev env setup can be a later task.

bartaz:
I think it's fine to merge it and use it as a base for improvement and work in new branches.

server/server.js
earnubs: I agree, I've seen server in ./src, maybe that's where it should go, but will leave for another PR

katiefenn:
👍

PR #5, opened 2016-08-16
Added normalize.css and simple base.css global stylesheets

bartaz:
normalize.css (linked from node module) and simple base.css added as global stylesheets.


katiefenn:
👍

PR #7, opened 2016-08-16
Coverage

earnubs:
Add coverage reporting via [istanbul nyc](https://github.com/istanbuljs/nyc) and [coveralls.io](https://coveralls.io/github/canonical-ols/javan-rhino)


package.json
```diff
@@ -7,8 +7,9 @@
   "scripts": {
     "start": "node server/",
     "lint": "eslint -c .eslintrc.js '**/*.js'",
-    "test": "npm run lint && mocha --compilers js:babel-register --recursive",
-    "test:watch": "npm test -- --watch"
+    "test": "npm run lint && nyc --require babel-core/register mocha --compilers js:babel-register --recursive",
```
katiefenn: Is there a missing ampersand between the `nyc` and `mocha` commands?

earnubs: I don't think so, nyc is running mocha rather than running before mocha; see the last code block, where nyc is running `npm test`: https://github.com/istanbuljs/nyc#instrumenting-your-code

PR #8, opened 2016-08-16
20: Add CSS Modules config to WebPack

katiefenn:
- Added ExtractTextPlugin dependency
- Added configuration for css-loader & modules
- Added stylesheet bundle file to page template

### DIFF
--- a/webpack/dev-config.js
+++ b/webpack/dev-config.js
@@ -1,5 +1,6 @@
-var path = require('path');
-var webpack = require('webpack');
+const path = require('path');
+const webpack = require('webpack');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 module.exports = {
   entry: [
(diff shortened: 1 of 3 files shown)


webpack/dev-config.js
```diff
@@ -25,7 +27,10 @@ module.exports = {
       },
       {
         test: /\.css$/,
-        loader: 'style-loader!css-loader'
+        loader: ExtractTextPlugin.extract(
```
katiefenn: Loader config based on [CSS Modules example config](https://github.com/css-modules/webpack-demo/blob/master/webpack.config.js#L21).

webpack/dev-config.js
```diff
@@ -1,5 +1,6 @@
-var path = require('path');
-var webpack = require('webpack');
+const path = require('path');
+const webpack = require('webpack');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
```
katiefenn: Added ExtractTextPlugin to bundle CSS into an external stylesheet.

As well as just being A Good Idea, the external stylesheet can be cached by browsers and reduces the footprint of components.

PR #9, opened 2016-08-16
Add react-router

earnubs:
Add basic routes with react-router, also on server side.

NB. testing here is not sufficient, but we need jsdom or something to get a DOM and be more rigorous in the tests, I've added a card on Trello.


src/routes.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Route, IndexRoute } from 'react-router';
+import App from './containers/app.js';
+import About from './containers/about.js';
+import Home from './containers/home.js';
+
+export default (
+  <Route path="/" component={App}>
```
bartaz: Writing routes in "HTML" ... that's something to get used to :)

server/server.js
```diff
@@ -1,10 +1,12 @@
 import Express from 'express';
 import React from 'react';
-import ReactDOM from 'react-dom/server';
+import {renderToString} from 'react-dom/server';
+import { match, RouterContext } from 'react-router';
```
bartaz: I noticed we are bit inconsistent in using brackets in imports (and other places).
I don't know if eslint can lint that. But maybe something to consider.

bartaz:
Looks good, and it's pre-rendered on server side! 👏

server/server.js
earnubs: Yep http://eslint.org/docs/rules/object-curly-spacing

My vote is { foo }, but I'll not break down in tears if we vote for {foo} ...

katiefenn: +1 { foo }

PR #11, opened 2016-08-17
move css to head, fixes #10

### LINKED ISSUE #10: CSS delay on page load
The initial page render has no style then it flicks to styled when the bundle loads, the <link /> for style should be moved into <head>

earnubs:
Tests to be added once enzyme lands

### DIFF
--- a/src/containers/app.js
+++ b/src/containers/app.js
@@ -1,10 +1,23 @@
 import React, { PropTypes, Component } from 'react';
 import NavLink from '../components/navlink';
+import Helmet from 'react-helmet';
 
 export default class App extends Component {
   render() {
+    // FIXME sstewart 17-Aug-16 move some of this to config
     return (
       <div>
+        <Helmet
+          htmlAttributes={{'lang': 'en'}}
+          title='buy [snap-name]'
+          titleTemplate='my.ubuntu.com - %s'
+          meta={[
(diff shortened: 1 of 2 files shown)


src/helpers/html.js
```diff
@@ -16,11 +16,11 @@ export default class Html extends Component {
           {head.title.toComponent()}
           {head.meta.toComponent()}
           {head.link.toComponent()}
+          <link rel="stylesheet" href="/static/style.css" />
```
bartaz: Should/could this be done with helmet? We have {head.link.toComponent()} right above it.

src/containers/app.js
```diff
@@ -1,10 +1,23 @@
 import React, { PropTypes, Component } from 'react';
 import NavLink from '../components/navlink';
+import Helmet from 'react-helmet';
 
 export default class App extends Component {
   render() {
+    // FIXME sstewart 17-Aug-16 move some of this to config
```
bartaz: Agreed that some of Helmet props may be better extracted to some config or to Html template, but it can be done later on.

bartaz:
Looks good to me. +1

src/containers/app.js
earnubs: I quickly looked at nconf, adding this (and card) as a reminder to look more later

PR #13, opened 2016-08-17
Added Enzyme to tests

bartaz:
Added enzyme and updated tests to use it.

Also added more tests of Html helper.
Testing with Helmet is bit tricky, but works.

Note: Running tests throws some React warnings due to a bug in Helmet.

### DIFF
--- a/test/helpers/t_html.js
+++ b/test/helpers/t_html.js
@@ -1,14 +1,47 @@
 import React from 'react';
 import expect from 'expect';
-import { shallow } from 'enzyme';
+import { render, shallow } from 'enzyme';
+import Helmet from 'react-helmet';
 
 import Html from '../../src/helpers/html.js';
 
+// NOTE:
+// due to a bug in react-helmet [1] React throws a warning while running tests
+// without Helmet passed to Html helper:
+//
+//    Warning: React.createElement(...): Expected props argument to be a plain object.
+//    Properties defined in its prototype chain will be ignored.
+//
+// [1] https://github.com/nfl/react-helmet/issues/158
 describe('<Html /> helper', () => {
 
   it('should be an html block', () => {
     const html = shallow(<Html />);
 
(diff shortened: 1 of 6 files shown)


test/helpers/t_html.js
```diff
@@ -1,14 +1,47 @@
...
-    const renderer = createRenderer();
-    const html = renderer.render(<Html />);
+    const html = shallow(<Html />);
+
+    expect(html.type()).toEqual('html');
+  });
+
+  it('should render content given as prop', () => {
+    const content = <span>test</span>;
+    // because of dangerouslySetInnerHTML we need to do static render instead of shallow
```
bartaz: Because we use `dangerouslySetInnerHTML` in `Html` I needed to use `render` instead of `shallow`.

I'm not sure if we really need `dangerouslySetInnerHTML` if we pass React component anyway and can render it on server side?

The only thing that seems to be needing `renderToString()` is Helmet, which requires it to work on server side. But we can still do `renderToString` just for Hammer, but use `{component}` instead of `dangerouslySetInnerHTML` to render content.

earnubs: This was taken from https://github.com/erikras/react-redux-universal-hot-example/blob/master/src/helpers/Html.js, and I'm not sure why `dangerouslySetInnerHTML` is used there. I quickly tried using `props.children` and it worked but I'll be honest and say I don't completely understand the side effects, see here: https://github.com/canonical-ols/javan-rhino/pull/15

katiefenn: I'm worried about using dangerouslySetInnerHTML in such a fundamental location. I think it would be easy to introduce a vulnerability by doing this.

bartaz: We all agree to that. Let's try to fix it in follow up branch and not block merging this (as it didn't introduce `dangerouslySetInnerHTML`).

@earnubs @katiefenn  Agreed?

PR #14, opened 2016-08-17
22: Add PostCSS

katiefenn:
- Added PostCSS pre-processor
- Added postcss-import plugin
- Added postcss-simple-vars plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,9 +50,10 @@
     "style-loader": "^0.13.1",
     "webpack": "^1.13.1",
     "webpack-dev-middleware": "^1.6.1",
-    "webpack-hot-middleware": "^2.12.2"
-  },
-  "devDependencies": {
-    "extract-text-webpack-plugin": "^1.0.1"
+    "webpack-hot-middleware": "^2.12.2",
+    "extract-text-webpack-plugin": "^1.0.1",
+    "postcss-import": "^8.1.2",
+    "postcss-loader": "^0.10.0",
+    "postcss-simple-vars": "^3.0.0"
   }
 }
--- a/webpack/dev-config.js
+++ b/webpack/dev-config.js
@@ -1,6 +1,17 @@
 const path = require('path');
 const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const atImport = require('postcss-import');
+const variables = require("postcss-simple-vars");
+
+const cssLoaderConfig = [
+  'css-loader',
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -50,9 +50,10 @@
     "style-loader": "^0.13.1",
     "webpack": "^1.13.1",
     "webpack-dev-middleware": "^1.6.1",
-    "webpack-hot-middleware": "^2.12.2"
-  },
-  "devDependencies": {
-    "extract-text-webpack-plugin": "^1.0.1"
```
katiefenn: Moved all the things plugged into WebPack into `dependencies` instead of `devDependencies`

webpack/dev-config.js
```diff
@@ -1,6 +1,8 @@
 const path = require('path');
 const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const atImport = require('postcss-import');
+const variables = require('postcss-simple-vars');
```
katiefenn: Added PostCSS plugins for Sass-esque loading of files and variables.

webpack/dev-config.js
```diff
@@ -28,10 +30,32 @@ module.exports = {
...
+          ].join('!'),
+          [
+            loader(
+              'css-loader',
+              'modules',
+              'importLoaders=1',
+              'localIdentName=[name]__[local]___[hash:base64:5]'
+            ),
+            'postcss-loader'
+          ].join('!')
```
katiefenn: Tried to make config more readable than the rubbish WebPack loader config strings.

Shout if this is over-the-top.

webpack/dev-config.js
```diff
@@ -28,10 +30,32 @@ module.exports = {
...
         )
       }
     ]
+  },
+  postcss: function () {
+    return [atImport, variables];
   }
 };
+
```
katiefenn: Helper function for making WebPack loader config strings.

package.json
earnubs: since we'll never publish this as a module, does it matter? (and then, why not have everything in deps )

katiefenn: On the last app I worked on we bundled our `dependencies` with the apps in a single JS file (excluding `devDependencies`). Moving mocha to devDependencies would reduce the response by 9K (unzipped) in filesize in this case.

Perhaps the best thing to do would be bung everything in `dependencies` and review when we actually intend to deploy?

PR #15, opened 2016-08-17
use props.children instead of dangerouslySetInnerHTML

earnubs:


src/helpers/html.js
```diff
@@ -1,13 +1,10 @@
 import React, { Component, PropTypes } from 'react';
-import ReactDOM from 'react-dom/server';
 import Helmet from 'react-helmet';
 
 export default class Html extends Component {
 
   render() {
 
-    const { component } = this.props;
-    const content = component ? ReactDOM.renderToString(component) : '';
```
bartaz: Removing this line will cause Helmet to fail in tests. If  `ReactDOM.renderToString(component)` (with Helmet inside) is not called Helmet will be empty (even if set in tests) and will not render anything - making it impossible to test anything related to Helmet output.

I tried different approaches (like using `render` from enzyme), but I couldn't find a workaround. I'm not sure if it's due to Helmet itself (they say in docs to call helmet after `renderToString`) or due to our way of rendering `<html>` in React, or combination of both.

PR #18, opened 2016-08-18
22: Add PostCSS

katiefenn:
- Added PostCSS pre-processor
- Added postcss-simple-vars plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "webpack-dev-middleware": "^1.6.1",
     "webpack-hot-middleware": "^2.12.2",
     "extract-text-webpack-plugin": "^1.0.1",
-    "postcss-import": "^8.1.2",
     "postcss-loader": "^0.10.0",
     "postcss-simple-vars": "^3.0.0"
   }
--- a/package.json
+++ b/package.json
@@ -52,9 +52,10 @@
     "style-loader": "^0.13.1",
     "webpack": "^1.13.1",
     "webpack-dev-middleware": "^1.6.1",
-    "webpack-hot-middleware": "^2.12.2"
-  },
-  "devDependencies": {
-    "extract-text-webpack-plugin": "^1.0.1"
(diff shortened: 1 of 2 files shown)


katiefenn:
I've removed the postcss-import plugin, because it seems too powerful to use with CSS Modules.

The original reason for including it was to use it to import variables from other files, like you can in Sass. But postcss-import also allows you to import rules from other files, which undermines the modularity of CSS Modules.

postcss-simple-vars allows the injection of variables into stylesheets, but this will be managed in configuration instead.

earnubs:
Could you add autoprefixer? Default setting should be fine I think?

bartaz:
So without postcss-import modules don't have any access to other CSS?

I wonder how it affects possibility of using [class composition](https://github.com/css-modules/css-modules#composition). For example having some shared base classes defined and modules using them to compose local styles?

But maybe I'm trying to solve the problem that we don't have (yet).

katiefenn:
bartaz: That's it exactly - I'd prefer to encourage using the class composition feature in CSS Modules instead of importing rules like we do in Sass. I'm concerned that doing so risks bloating the CSS bundle file.

bartaz:
@katiefenn So can you compose from a class outside of current module without importing it? Assuming it will be global maybe? I haven't used that, so I'm not sure how it works in practice. Maybe it will be clearer once we have some components code to work with.

katiefenn:
bartaz: Yes. It works differently to Sass's global importing.

Say `.rule-a` composes `.rule-b`: wherever the `.rule-a` className is added to an element, CSS Modules will automatically add the `.rule-b` className too, ensuring that the element inherits the styles from both rules.

earnubs:
🛳

bartaz:
@katiefenn Right! I knew that it adds class names to elements, but for some reason still thought it requires importing of some sort. Yay CSS Modules! 🎉

PR #22, opened 2016-08-24
30: Fix server-side rendering and CSS Modules

katiefenn:
- Install webpack-isomorphic-tools
- Separate starting WebPack dev server from starting Node server
- Add npm script for starting WebPack dev server
- Add dynamic assets config to server page template

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,10 @@
   "description": "Front-end to Ubuntu Store Payments",
   "main": "index.js",
   "scripts": {
-    "start": "node server/",
+    "start": "concurrently --kill-others \"npm run watch-client\" \"node server/\"",
+    "watch-client": "node webpack/webpack-dev-server.js",
     "lint": "eslint -c .eslintrc.js '**/*.js'",
-    "test": "npm run lint && nyc --require babel-core/register mocha --compilers js:babel-register --recursive",
+    "test": "nyc --require babel-core/register mocha --compilers js:babel-register --recursive && npm run lint",
     "test:watch": "npm test -- --watch",
     "coverage": "nyc report --reporter=text-lcov | coveralls"
   },
@@ -34,6 +35,7 @@
     "babel-preset-react": "^6.11.1",
     "babel-preset-stage-0": "^6.5.0",
     "babel-register": "^6.11.6",
+    "concurrently": "^2.2.0",
     "coveralls": "^2.11.12",
     "css-loader": "^0.23.1",
     "enzyme": "^2.4.1",
@@ -56,6 +58,7 @@
     "style-loader": "^0.13.1",
     "webpack": "^1.13.1",
     "webpack-dev-middleware": "^1.6.1",
-    "webpack-hot-middleware": "^2.12.2"
+    "webpack-hot-middleware": "^2.12.2",
+    "webpack-isomorphic-tools": "^2.5.7"
   }
 }
--- a/package.json
+++ b/package.json
@@ -30,6 +30,8 @@
     "autoprefixer": "^6.4.0",
     "babel-core": "^6.13.2",
     "babel-loader": "^6.2.4",
+    "babel-plugin-transform-es2015-destructuring": "^6.9.0",
+    "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-polyfill": "^6.13.0",
     "babel-preset-es2015": "^6.13.2",
     "babel-preset-react": "^6.11.1",
--- a/server/server.js
+++ b/server/server.js
@@ -8,7 +8,7 @@ import Html from '../src/helpers/Html';
 
 const app = Express();
 
-module.exports = webpackIsomorphicTools => {
+module.exports = (webpackIsomorphicTools) => {
   app.use('/', function (req, res) {
 
     if (process.env.NODE_ENV !== 'production') {
--- a/server/server.js
+++ b/server/server.js
@@ -38,11 +38,12 @@ module.exports = webpackIsomorphicTools => {
   });
 
   const port = process.env.PORT || 3000;
+  const address = process.env.ADDRESS || 'localhost';
 
-  const server = app.listen(port, () => {
+  const server = app.listen(port, address, () => {
     const host = server.address().address;
     const port = server.address().port;
 
-    util.log('try-auth app listening at http://%s:%s', host, port);
+    util.log('🚂  Express server listening on http://%s:%s 🚂', host, port);
   });
 };
--- a/server/server.js
+++ b/server/server.js
@@ -3,52 +3,46 @@ import React from 'react';
 import { renderToString } from 'react-dom/server';
 import { match, RouterContext } from 'react-router';
 import util from 'util';
-import webpack from 'webpack';
-import webpackDevMiddleware from 'webpack-dev-middleware';
-import webpackHotMiddleware from 'webpack-hot-middleware';
 import routes from '../src/routes';
 import Html from '../src/helpers/Html';
-import webpackConfig from '../webpack/dev-config';
 
 const app = Express();
-const compiler = webpack(webpackConfig);
 
-app.use(webpackDevMiddleware(compiler, {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -5,9 +5,10 @@
   "description": "Front-end to Ubuntu Store Payments",
   "main": "index.js",
   "scripts": {
-    "start": "node server/",
+    "start-dev": "concurrently --kill-others \"npm run watch-client\" \"node server/\"",
```
katiefenn: Renamed start script to make it clear site is starting with development settings.

package.json
```diff
@@ -5,9 +5,10 @@
   "description": "Front-end to Ubuntu Store Payments",
   "main": "index.js",
   "scripts": {
-    "start": "node server/",
+    "start": "concurrently --kill-others \"npm run watch-client\" \"node server/\"",
```
katiefenn: webpack-isomorphic-tools requires WebPack server to be run in advance of, or concurrently with, the server script.

package.json
```diff
@@ -5,9 +5,10 @@
...
+    "watch-client": "node webpack/webpack-dev-server.js",
```
katiefenn: New, separate task to start WebPack development server and watch files.

package.json
```diff
@@ -5,9 +5,10 @@
...
     "lint": "eslint -c .eslintrc.js '**/*.js'",
-    "test": "npm run lint && nyc --require babel-core/register mocha --compilers js:babel-register --recursive",
+    "test": "nyc --require babel-core/register mocha --compilers js:babel-register --recursive && npm run lint",
```
katiefenn: **[UNRELATED]** Would like to swap test and lint tasks around, because inserting debugging statements into files while testing causes linting to fail, which means unit tests will never be run. It also feels like the correct order of priority for assessing with code quality.

server/server.js
```diff
@@ -3,52 +3,46 @@ import React from 'react';
...
-      // below, if you're using a catch-all route.
-      const component = <RouterContext {...renderProps} />;
-
-      res.status(200);
-      res.send('<!doctype html>\n' +
-      renderToString(<Html component={component} />));
-    } else {
-      res.status(404).send('Not found');
+
+module.exports = webpackIsomorphicTools => {
```
katiefenn: Wrapped server config inside a function so that `webpackIsomorphicTools` dependency can be passed in, instead of [using a global](https://github.com/erikras/react-redux-universal-hot-example/blob/master/bin/server.js#L24) like react-redux-universal-hot-example.

src/containers/app.js
```diff
@@ -14,9 +14,6 @@ export default class App extends Component {
           meta={[
             { 'name': 'description', 'content': 'my.ubuntu.com payments ui' },
           ]}
-          link={[
-            { 'rel': 'stylesheet', 'href': '/static/style.css' }
-          ]}
```
katiefenn: Removed this until I can figure out how to pass in stylesheet path from webpack-isomorphic-tools into the `App` component.

This is changing because `href` will change depending on the environment. When in development, it will be served by the WebPack development server, running on port 3001. When in production, it will be served as a static asset from the same server.

Alternatively, we can hardcode the port here and worry about how it works in production later.

webpack/dev-config.js
```diff
@@ -1,24 +1,31 @@
...
 
+const webpackIsomorphicToolsPlugin =
+  new WebpackIsomorphicToolsPlugin(require('./webpack-isomorphic-tools-configuration'))
+  .development();
+
 module.exports = {
+  context: path.resolve(__dirname, '..'),
   entry: [
-    'webpack-hot-middleware/client',
+    'webpack-hot-middleware/client?path=http://localhost:3001/__webpack_hmr',
```
katiefenn: Changed port to reflect that WebPack development server now runs on a separate port to the Express server.

webpack/webpack-dev-server.js
```diff
@@ -0,0 +1,32 @@
+require('babel-register');
```
katiefenn: New file for starting WebPack development server separately from Express server.

package.json
bartaz: I'm fine with that. I found lint failing during development or debugging quite annoying too.

If you run `test:watch` it only re-runs tests, so it's also a way to skip lint while developing.

server/server.js
bartaz: It took me a little while to see what happens in this line (my eyes not used to ES6 magic).

I find declaring arrow functions without parens a bit confusing.
Are you fine @katiefenn @earnubs with adding a code style rule to always require parens around arguments of arrow functions? http://eslint.org/docs/rules/arrow-parens
It makes things more explicit and easier to read IMHO.

src/containers/app.js
bartaz: @katiefenn In theory it doesn't have to be `App` that sets link (or other things on Helmet). So it should be possible to import Helmet and set link elsewhere, and it should render it in Html.

But I'm not 100% sure how that would work if Helmet is called outside of React components in server side code.

webpack/webpack-dev-server.js
```diff
@@ -0,0 +1,32 @@
...
+
+app.use(Express.static('public'));
+
+var port = process.env.PORT || 3001;
+
+var server = app.listen(port, () => {
+  var host = server.address().address;
+  var port = server.address().port;
+
+  util.log('try-auth app listening at http://%s:%s', host, port);
```
bartaz: I guess the `try-auth app` message is relict from the past that gets copied around :)

bartaz:
Looking good. Great work! Let's finally try some of that CSS modules magic :)

package.json
earnubs: Another annoyance is lint complaining about console statements on the server side, where they are really fine.

server/server.js
earnubs: +1 if it helps, I'm starting to get used to it though and think it's quite nice :)

webpack/webpack-dev-server.js
earnubs: I think `babel-register` is starting to get in the way a little on the server side, I can't try things out from cli without babel-node, and they're noticeably slower ... so, server side only stuff might be better using `require` rather than `import` and dumping `babel-register`, even at this early stage.

server/server.js
katiefenn: Happy to always use parens. I agree it makes arrow function syntax confusing to learn. 👍

src/containers/app.js
katiefenn: bartaz: Oh, cool. I'm sure I can find a way, in that case.

package.json
bartaz: Agreed on server side console.
But client side console logging would be useful for development/debugging too. Maybe it would be worth adding some simple logger that would be disabled on prod?

webpack/webpack-dev-server.js
earnubs: ... but we can tackle this in another branch

package.json
katiefenn: bartaz: Agreed. I have some idea of what a logging utility should look like from my last place.

package.json
```diff
@@ -29,11 +30,14 @@
     "autoprefixer": "^6.4.0",
     "babel-core": "^6.13.2",
     "babel-loader": "^6.2.4",
+    "babel-plugin-transform-es2015-destructuring": "^6.9.0",
+    "babel-plugin-transform-object-rest-spread": "^6.8.0",
```
katiefenn: Installed missing Babel plugins

webpack/dev-config.js
```diff
@@ -59,7 +61,7 @@ module.exports = {
   }
 };
 
-function loader(name, ...options) {
+function loader(name, options) {
```
katiefenn: Removing need for spread operator unsupported in Node.JS 4.

package.json
earnubs: fixes #23

PR #28, opened 2016-08-29
Header component

bartaz:
Added static header component with some basic styles.

my-ubuntu-com-header-static

@earnubs @katiefenn Some open questions in inline comments.


server/server.js
```diff
@@ -9,6 +9,8 @@ import Html from '../src/helpers/Html';
 const app = Express();
 
 module.exports = (webpackIsomorphicTools) => {
+  app.use(Express.static('public'));
```
bartaz: Adding serving static files from `/public` to be able to serve images (like Ubuntu logo).

@kfenn - I guess this may affect your branch with prod build and environment.

src/components/navlink/navlink.js
```diff
@@ -0,0 +1,17 @@
+import React, { PropTypes, Component } from 'react';
+import { Link } from 'react-router';
+
+import styles from './navlink.css';
+
+export default class NavLink extends Component {
+  render() {
+    let { isLast, className, ...other } = this.props;
+
+    return <Link {...other} className={ `${className} ${isLast ? styles.lastLink : styles.link}` } activeClassName={ styles.active } />;
```
bartaz: This is a bit strange way of building `className`, because it combines both className we can get from parent component (`Header` in this case) with local styles.

I'm not sure if that's a good practice and how to best handle it with CSS modules.

Should it be possible to set className from parent? Or should all styles available for given component be local only and any modifiers should be done with some custom props (like `isLast` in this example)?

src/components/header/header.css
```diff
@@ -0,0 +1,28 @@
+.header {
+  background-color: #fff;
```
bartaz: How much of that we want to keep in variables?
How many of them should be shared (via `variables.js` - making component and CSS module dependent on these values) and how many just local (possibly duplicated between components?

src/components/header/header.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from 'react';
+import { Link } from 'react-router';
+import NavLink from '../navlink/navlink';
```
bartaz: Having components in separate directories makes project files better organised, but it leads to such double names in imports. Not a big issue, but looks bit strange.

One workaround would be to name main script of component in folder `index.js` (and simply do `import NavLink from '../navlink'`), but this will lead to every component having its own index.js making it really hard to find things by file name.

Any opinions?

test/components/t_header.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
bartaz: As we have all component files together in separate folder in src maybe we could move tests there too?

src/components/navlink/navlink.js
earnubs: Wouldn't `isLast` class be something CSS should handle with :last-child or similar?

Also, reminds me of https://github.com/JedWatson/classnames, not that I'm suggesting we use it for this one case.

server/server.js
katiefenn: Yeah, although I've already got rebasing to do, so it's not a problem 😉

src/components/header/header.css
earnubs: I am always conflicted about variables. `#fff` is perfectly clear, so is `#000`, `#dd4814` is clear only once you've been here a while but possible to make mistakes; so maybe we have a very small set of vars for the more complex values only?

src/components/header/header.js
earnubs: Other suggestions: https://gist.github.com/branneman/8048520

public/img/ubuntu-logo.svg
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
```
earnubs: FWIW this SVG can be compressed a little (20%) https://pastebin.canonical.com/164252/ https://jakearchibald.github.io/svgomg/

bartaz: Cool, will do, thanks. Just copied what we have in SCA.

src/components/navlink/navlink.js
bartaz: @earnubs Yes, `isLast` could be handled by last-child, but that will mean depending on a markup - component that uses it will need to 'know' to wrap set of links in some parent element so that the last one is really last.

`classnames` is good suggestion. But also here is other issue (or rather coding practice). Should parent component be allowed to set `className` on child component directly or should only component itself set its own `className`.

public/img/ubuntu-logo.svg
katiefenn: Are we expecting to use this logo anywhere else? This could live in the component directory if not.

src/components/header/header.css
katiefenn: Agreed. Put the few, most common colours in variables and we'll see how far that gets us.

src/components/header/header.js
katiefenn: I'm in favour of using an `index.js` as an entry-point for the module. This would work well for single-component modules and multi-component modules:

`import { NavLink } from '../navlink';`
`import { InputText } from '../vanilla-forms';`
`import { InputRadio } from '../vanilla-forms';`

src/components/navlink/navlink.css
```diff
@@ -0,0 +1,28 @@
+$height: 50px;
+$link-color: $ubuntu-orange;
+$border-color: #ccc;
+
+.link {
+  box-sizing: border-box;
+  display: inline-block;
+  height: $height;
+  line-height: $height;
```
katiefenn: picky: Making `$height` a variable feels unnecessary. It might be more readable just keeping it in-line.

src/components/navlink/navlink.js
katiefenn: @bartaz: I think the parent shouldn't pass in classes directly. If there is a need for position-specific styles, there should be an `last` prop on the component.

PR #30, opened 2016-08-31
Added basic static footer component

bartaz:
Simple footer. Design based on SnapWeb and SCA for consistency:

my-ubuntu-footer-static

### DIFF
--- a/src/components/footer/index.js
+++ b/src/components/footer/index.js
@@ -1,0 +1,16 @@
+import React, { Component } from 'react';
+
+import styles from './footer.css';
+
+export default class Footer extends Component {
+  render() {
+    return (
+      <div className={ styles.footer }>
+        <div className={ styles.container }>
+          <p className={ styles.copyright }>© 2016 Canonical Ltd. Ubuntu and Canonical are registered trademarks of Canonical Ltd.</p>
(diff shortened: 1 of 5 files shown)


src/style/base.css
```diff
@@ -56,6 +56,11 @@ a:focus {
   outline: thin dotted;
 }
 
+html {
+  min-height: 100%;
+  position: relative;
```
bartaz: Needed to keep absolute positioned footer at the bottom.

src/components/footer/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+import styles from './footer.css';
+
+export default class Footer extends Component {
+  render() {
+    return (
+      <div className={ styles.footer }>
+        <div className={ styles.container }>
+          <p className={ styles.copyright }>© 2016 Canonical Ltd. Ubuntu and Canonical are registered trademarks of Canonical Ltd.</p>
+          <p><a href="#">Terms of Service</a> · <a href="https://github.com/canonical-ols/javan-rhino/issues/new">Report a bug on this site</a></p>
```
bartaz: We need to define what links we want in footer and where should they point to.

earnubs: I'd expect the TOS to be on my.ubuntu.com, for content we'll need to raise that with @evandandrea 

Report a bug can point to github issues for now.

bartaz: Added a card to Trello about that.

PR #32, opened 2016-08-31
33: Return app from server for tests

katiefenn:
Sorry, poor communication on my part, I've done this in my own branch. I was asking just if there was any reason not to!

Also, needs to return app but not run server, so I exported app itself rather than return app from the default export.


bartaz:
LGTM 👍

PR #35, opened 2016-09-01
Don't hardcode localhost.

evandandrea:


earnubs:
This will be handled with nconf in auth pr

PR #36, opened 2016-09-01
ubuntu one authentication

earnubs:
OpenID dance for mu, with nconf for configuring the app. I would expect applicable client side nconf config to be rendered on the server into __INITIAL_STATE__ and hydrated to redux state, thus making the config universal, but we'll keep that part for another branch.

Tests depend on `mongod` running (see README), install and starting of mongod should be part of another PR. We'll split out these tests and so some kind of feature detection in tests in another branch.


.travis.yml
```diff
@@ -2,4 +2,6 @@ language: node_js
 node_js:
   - "6"
```
earnubs: If this PR is dropping node to v4, then this needs to change too

test/handlers/t_login.js
```diff
@@ -0,0 +1,89 @@
...
+        expect(error).toEqual({
+          message:
+            'No providers found for the given identifier'
+        });
+      });
+    });
+
+    describe('authUrl query string', () => {
+      it('should have the correct realm', () => {
+        const rp = createRelyingParty();
```
katiefenn: Can some of this setup be added to a `beforeEach` hook?

test/handlers/t_login.js
```diff
@@ -0,0 +1,39 @@
...
+    const req = {
+      session: {
+        destroy: spy
+      }
+    };
+    const res = {};
+
+    logout(req, res);
+
+    expect(spy.calls.length).toEqual(1);
```
katiefenn: Do these assertions test the same thing? Do we need both?

test/handlers/t_login.js
```diff
@@ -0,0 +1,39 @@
...
+    expect(spy).toHaveBeenCalled();
+  });
+
+  it('logout redirects to home if destroy successful', () => {
+    const req = {
+      session: {
```
katiefenn: We could use [Sinon](http://sinonjs.org) to create a [stub](http://sinonjs.org) to call the callback for us. Sinon stubs are also spies, so a common `beforeEach` function could be created for this and the preceding test.

```
destroy: sinon.stub().callsArg(0);
```

...or something like this.

katiefenn: Picky: Also, I'd like to take this opportunity to  cheer-lead for skinny `it` blocks, following [this style](https://medium.com/@lewiscowper/test-all-the-things-e0d0f03b382e#.givnie335).

package.json
```diff
@@ -72,6 +72,7 @@
     "react-hot-loader": "^1.3.0",
```
katiefenn: What's itses, Precious?

Picture of Gollum

server/server.js
```diff
@@ -1,17 +1,34 @@
...
+import conf from './configure.js';
+import login from './routes/login';
 import routes from '../src/routes';
-import Html from '../src/helpers/Html';
+import util from 'util';
+import { match, RouterContext } from 'react-router';
+import { renderToString } from 'react-dom/server';
+
+const MongoStore = MongoConnect(session);
+const sesssionStore = new MongoStore({
```
katiefenn: Typo

test/handlers/t_login.js
```diff
@@ -0,0 +1,51 @@
...
+import { spy, stub } from 'sinon';
+import { logout } from '../../server/handlers/login.js';
+
+describe('login handler', () => {
+
+  // session.destroy stub
+  let destroy;
+
+  const req = {
+    session: {
```
katiefenn: Sinon can do this for us: 
```
stub.callsArg(0)
```

PR #38, opened 2016-09-02
Payment details form

bartaz:
Static payments form to review.

Created a bunch of simple form component to build the form.
I guess more encapsulation will follow when special validation (such as credit card) will show up.


src/components/forms/checkboxField.js
```diff
@@ -0,0 +1,19 @@
...
+
+import Label from './label';
+
+import style from './forms.css';
+
+export default function CheckboxField(props) {
+  const { id, label, ...other } = props;
+
+  return <Label htmlFor={ id }>
+    <input type="checkbox" id={ id } {...other} className={ style.tickInput } />
```
katiefenn: I think we should avoid passing in arbitrary props into components. It undermines the benefits of encapsulation.

src/components/payments-form/index.js
```diff
@@ -0,0 +1,59 @@
...
+      <Form>
+        <h3>Payment details</h3>
+
+        <Fieldset>
+          <InputField id="customerCountry" label="Country" placeholder="United Kingdom" /> {/* TODO: select? */}
+        </Fieldset>
+
+        <Fieldset>
+          <h4>Name and address</h4>
+          <InputField id="customerFullname" label="Full name" placeholder="John Doe" />
```
katiefenn: Do fields need IDs if we're not going to be traversing the DOM, jQuery style?

src/components/forms/forms.css
```diff
@@ -0,0 +1,39 @@
+.fieldset {
```
katiefenn: These styles could be split out into files for each input type. If you need to share styles, you could try using `composes`.

src/components/payments-form/countries.js
```diff
@@ -0,0 +1,251 @@
+export default [
+{ iso: 'AF', name: 'Afghanistan' },
```
bartaz: Countries list with ISO codes as we have in SCA.

src/components/forms/selectField.js
```diff
@@ -0,0 +1,28 @@
+import React, { PropTypes } from 'react';
+
+import Label from './label';
+import Select from './select';
+
+import styles from './inputField.css';
+
+export default function SelectField(props) {
```
bartaz: This is quite similar to InputField (difference is `Input` vs `Select` as a child input), so I wonder if it would be useful to have them share implementation in some more generic `Field` component?

src/components/forms/select.css
```diff
@@ -0,0 +1,19 @@
+$color: $mid-grey;
+
+.select {
+  position: relative;
+}
+
+.select:after {
+  content: "∨";
```
bartaz: Quick workaround to make select look similar to the designs without using any images for arrow.

src/components/forms/form.css
```diff
@@ -0,0 +1,2 @@
+.form {
```
bartaz: Empty placeholder for now, I expect some styles to be added later.

I wonder if the width of the form would fit here better then in `fieldset.css`?

src/style/base.css
```diff
@@ -56,11 +56,51 @@ a:focus {
...
-html {
-  min-height: 100%;
-  position: relative;
-}
-
 body {
   background-color: $light-grey;
 }
+
+h1,
```
bartaz: Heading sizes based on vanilla CSS.

src/components/button/button.css
```diff
@@ -0,0 +1,50 @@
+$primary-button-color: $ubuntu-orange;
+$primary-button-color-hover: #d44615;
+$primary-button-text-color: #FFF;
+
+$secondary-button-color: #FFF;
+$secondary-button-color-hover: #EFEFEF;
+$secondary-button-text-color: $ubuntu-orange;
+$secondary-border-color: $mid-grey;
+
+.button {
```
bartaz: Button styles based on vanilla CSS and SCA as our designs seem to have something in between ;)

src/components/button/index.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default function Button(props) {
+  const { type='primary' } = props;
+  return <button {...props} className={ style[type] } />;
+}
+
+Button.propTypes = {
+  children: PropTypes.string,
+  onClick: PropTypes.func,
+  type: React.PropTypes.oneOf(['primary', 'secondary'])
```
katiefenn: Isn't `type` a reserved [attribute of the button element](https://developer.mozilla.org/en/docs/Web/HTML/Element/button#attr-type)?

For this reason I really think we should at least whitelist some of the props we're passing in to the `button` component. Something like this could be good?

```
const { appearance = "primary", onClick, children }  = props;
return <button appearance={ appearance } onClick={ onClick }>{ children }</button>;
```

src/components/payments-form/countries.js
earnubs: Could we use something like https://github.com/michaelwittig/node-i18n-iso-countries instead? Or do we need control of this list?

src/components/forms/selectField.js
earnubs: My 2c is keep it as it is and then take a look again post GA.

src/components/forms/select.css
earnubs: Or ::before and ::after 1px x ~10px boxes rotated to form the arrow? :D

src/components/button/button.css
earnubs: For the purposes of this project, SCA style does not exist, only Vanilla; if you have to choose, choose not to make it look like SCA, if that makes sense?

earnubs: Lowercase hex values? Or am I alone now in my pedantry?

src/style/base.css
```diff
@@ -56,11 +56,51 @@ a:focus {
...
+h3,
+h4,
+h5,
+h6 {
+  font-weight: 300;
+  line-height: 1.3;
+}
+
+h1 {
+  font-size: 2.8125em;
```
earnubs: If this is Vanilla, fine, maybe I should suck it up, but I really don't like em values to calculate pixel values ...

src/components/button/button.css
bartaz: @earnubs In buttons case - styles are mostly vanilla, the only difference is in padding - buttons in Vanilla are bigger (more padding) then what's on the design, so SCA padding was much closer.

bartaz: @earnubs @katiefenn I'm on 'uppercase hex values' side, but when it comes to pedantry I choose consistency over my preference, so I'm fine with whatever we choose to use around the project(s).

src/components/footer/footer.css
```diff
@@ -1,5 +1,15 @@
 $border-color: #e3e3e3;
 
+/* html and body styles needed for footer positioning */
```
katiefenn: We should avoid introducing global styles in modular CSS. This could lead to clashes with other components.

Use case: Imagine if we prepared a style guide for our presentational components. Would we want these styles applied to the global styles outside of our application?

src/components/button/index.js
bartaz: Good spot. I didn't want to use `type` as HTML param (did it by mistake), only as our custom param for "appearance", but I guess with a generic `Button` component we may want to use `type` in HTML at some point as valid param, so I'll rename custom prop `appearance` and not include it in HTML as it's not needed there.

src/components/payments-form/countries.js
bartaz: @earnubs I looked into some modules with country names/codes (as there are couple of them of course), but it seemed simpler to use the one we already have and keep it as array already and not worry about getting alphabetically sorted list of all countries.

src/components/button/index.js
```diff
@@ -0,0 +1,14 @@
+import React, { PropTypes } from 'react';
```
katiefenn: Shouldn't this be in the forms directory?

src/components/forms/form.css
katiefenn: Hmm. I don't think so - forms come in all shapes and sizes. I believe `form` elements are functional rather than presentational - they determine what a form does rather than how a form looks.

src/components/forms/input.js
```diff
@@ -0,0 +1,14 @@
+import React, { PropTypes } from 'react';
+
+import style from './input.css';
+
+export default function Input(props) {
+  return <input {...props} className={ style.textInput }/>;
+}
+
+Input.propTypes = {
+  id: PropTypes.string,
```
katiefenn: id and name could be generated from a single prop.

src/components/forms/input.js
```diff
@@ -0,0 +1,14 @@
...
+  name: PropTypes.string,
+  placeholder: PropTypes.string,
+  type: PropTypes.string
```
katiefenn: We should make a separate component for each type of text element ("text", "password", "number", "tel"):

- We could have different validation rules and styles for each one.
- The range of input types would be clear and obvious, because they would be clearly documented.

src/components/forms/input.js
```diff
@@ -0,0 +1,14 @@
+import React, { PropTypes } from 'react';
```
katiefenn: It might be nice to create different components for each of the text field types we are expecting ("text", "id", "password", "tel")

src/components/forms/input.js
```diff
@@ -0,0 +1,14 @@
...
+
+import style from './input.css';
+
+export default function Input(props) {
+  return <input {...props} className={ style.textInput }/>;
```
katiefenn: Suggest whitelisting props.

src/components/forms/input.js
katiefenn: I wonder what this component is giving us it functionally mirrors the built-in `<input>` components.

src/components/forms/selectField.js
katiefenn: Thinking a bit further down the line, we might want fields with multiple inputs. For example, we might want a mutli-checkbox field with specific layout rules that would benefit from having specialised field components.

src/components/payments-form/countries.js
katiefenn: Could we encapsulate this within a `<CountryInput />` field?

src/components/footer/footer.css
bartaz: @katiefenn Good point, I was also thinking about this. On the other hand, these styles are only needed on html/body because of footer (for example explicit bottom margin is based on footer height), so having them in global styles also doesn't seem ok.

One other solution I see is having footer module define these styles as classes and footer module can apply these to html/body elements when it's rendered:

```css
.hasFixedFooter {
   min-height: 100%;
   position: relative;
 }
 
 .hasFooter {
   margin-bottom: 75px;
 }
 ```

```js
document.rootElement.classList.add(styles.hasFixedFooter);
document.body.classList.add(styles.hasFooter);
```

But this has kind of same result - footer module adding some styles to html/body, so I'm not sure if it changes anything.

Any other suggestion how to tackle this?

src/components/button/index.js
bartaz: I don't think that buttons are specific to forms. At some point we may want some navigation buttons, or other 'general purpose' buttons. On designs we already have some big buttons for sign in and completing purchase (in future), so `Button` component may get less specialised.

src/components/forms/input.js
bartaz: Encapsulates `<input>` with styles.

src/components/forms/selectField.js
bartaz: We are not there at the moment. I guess we can improve our form component as we need.
For sure adding validation will create opportunity to build some more specialised component for custom field types. And we can add more when we need to.

src/components/payments-form/countries.js
bartaz: Good point about encapsulating it in custom field. Will do it when adding validation.

src/components/footer/footer.css
katiefenn:
```
document.rootElement.classList.add(styles.hasFixedFooter);
document.body.classList.add(styles.hasFooter);
```
Yeah, this is even worse. You could always use the CSS Modules `@global` keyword, but I still think a fixed footer is a property of the page rather than the footer itself.

We could have a re-usable `<Page />` component to encapsulate header and footer management, perhaps? Or build it into `<App />`?

src/components/button/index.js
katiefenn: I wasn't thinking about where `<button />` appears in the DOM - of course a button can appear outside of a `<form />`. But if we were to package up our input elements for distribution on npm, I imagine them all living together.

src/components/forms/input.js
katiefenn: If we apply styles to an input element inside `<InputField />`, they'd still be encapsulated.

We _could_ implement the re-use of this code at the stylesheet level as well:

```
/* inputfield.css */
.input {
  composes: textInput from 'input.css';
}
```

src/components/footer/footer.css
bartaz: Even though this body bottom margin in theory could be moved to some other wrapper in `App` for example, still `Page` or `App` would have to add `min-height: 100%` to `<html>`.

This change is not even directly related to this PR (I just added these styles to fix footer broken for page with long content). So I'm raising it as an issue #44 to be fixed later.

PR #39, opened 2016-09-05
39: Add Redux

katiefenn:
- Installed Redux and related packages
- Created auth actions and reducer
- Added example login event handling (to be replaced by concrete implementation)


.babelrc
```diff
@@ -1,6 +1,7 @@
 {
   "presets": ["es2015", "stage-0", "react"],
   "plugins": [
+    "transform-decorators-legacy",
```
katiefenn: Added decorator plugin to help connect components to Redux stores. I'm not entirely comfortable with supporting ES7 syntax - I'd prefer to wait until the spec is finalised. Decorators are just too nice to not use for connecting components to stores 😐

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
       'eslint:recommended',
       'plugin:react/recommended'
     ],
+    "parser": "babel-eslint",
```
katiefenn: Added `babel-eslint` to because ESLint doesn't support ES7 decorators.

.eslintrc.js
```diff
@@ -23,7 +24,10 @@ module.exports = {
     'rules': {
         'indent': [
             'error',
-            2
+            2,
+            {
+              'SwitchCase': 1
+            }
```
katiefenn: Added new setting to correctly lint switch statement indentation.

src/components/header/index.js
```diff
@@ -20,10 +22,25 @@ export default class Header extends Component {
             <NavLink to="/about">About</NavLink>
           </div>
           <div className={ styles.sideNav }>
-            <NavLink to="/login">Login</NavLink>
+            {!loggedIn &&
```
katiefenn: Contrived example to illustrate how to hook components into stores. I'm sure we'll want to improve this code in the near-future when we get around to it.

src/redux/auth.js
```diff
@@ -0,0 +1,30 @@
+const INITIATE_LOGIN = 'auth/INITIATE_LOGIN';
+const COMPLETE_LOGIN = 'auth/COMPLETE_LOGIN';
+const FAIL_LOGIN = 'auth/FAIL_LOGIN';
+
+const initialState = {
+  loggedIn: false
+};
+
+export default function reducer(state = initialState, action = {}) {
+  switch(action.type) {
```
katiefenn: Only `COMPLETE_LOGIN` is used at the moment, I filled in some other actions that we'll probably use soon.

test/containers/t_app.js
```diff
@@ -1,15 +1,21 @@
...
+const mockStore = configureStore();
 
 describe('<App /> container', () => {
+  const store = mockStore({
+    auth: {
+      loggedIn: true
+    }
+  });
 
-  it('should render a div', () => {
```
katiefenn: This test doesn't work any more, I think it _might_ be because the component is wrapped by `connect`. I'm not motivated to fix it because it tests implementation and not anything of business value, and I assume it was more of a placeholder anyway.

test/helpers/t_html.js
```diff
@@ -45,11 +36,4 @@ describe('<Html /> helper', () => {
     expect(html.find('body link[rel="stylesheet"]').length).toEqual(0);
     expect(html.find('head link[rel="stylesheet"]').length).toEqual(1);
   });
-
-  it('should render title from Helmet', () => {
-    const helmet = <Helmet title='test title' />;
-    const html = shallow(<Html component={ helmet } assets={ assets } />);
-
-    expect(html.find('title').text()).toEqual('test title');
-  });
```
katiefenn: I've removed this test because it also doesn't work anymore. This is because `Helmet` is now not a direct descendant child of `Html` and hence not rendered in shallow-rendering.

Besides this, this test did not test anything that was not already tested on line 23, testing that children are rendered.

test/helpers/t_html.js
```diff
@@ -1,18 +1,9 @@
 import React from 'react';
 import expect from 'expect';
 import { render, shallow } from 'enzyme';
-import Helmet from 'react-helmet';
 
 import Html from '../../src/helpers/html.js';
 
-// NOTE:
```
katiefenn: Removed this note because Helmet is not rendered in shallow-rendering and this issue does not occur.

earnubs:
No store on the server side?

earnubs:
The login example is interesting and I know it's an exception here in real world usage, because the openid dance means redirects they'll be pages loaded from the server and never client side, which means that while the state items you've defined are good, I don't think we'll be setting them with actions, they'd come from some initial state rendered into the page? Is that correct?

test/helpers/t_html.js
bartaz: OK. Helmet and its 'global' nature was messing up in tests a little bit, so I'm fine with getting rid of it.

test/containers/t_app.js
bartaz: Yep, most of the tests we have currently are placeholders just checking type of rendered element.
It's good to update them to something more reasonable in branches like this one.

.eslintrc.js
bartaz: I like my `case`s indented too :) I was pretty surprised first time I saw eslint complaining about that.

src/containers/app.js
```diff
@@ -1,8 +1,14 @@
 import React, { PropTypes, Component } from 'react';
 import Helmet from 'react-helmet';
+import { connect } from 'react-redux';
 
 import Header from '../components/header';
 import Footer from '../components/footer';
+import * as authActions from '../redux/auth';
+
+@connect(
```
bartaz: That's something I need to get my head around. Not only this seems to be the place where react-redux magic happens, but the decorator adds another layer of confusion for me. 

I agree that declarative nature of decorators is nice, but my eyes are not used to see such things in JS code.

Standard:

```js
export default connect((state) => ({ loggedIn: state.auth.loggedIn }), authActions)(App)
```
is a bit more explicit (especially about wrapping the class) but not less magic at all.

I have a slight preference to not using decorators, but I can get used to them if we decide to give it a try.

bartaz:
I guess this may not be an issue yet, but at some point we may need to share the state/store between server and client, so that rendering on client can pick up data used by server to render.

katiefenn:
@earnubs: The store should be added to the server-side rendering [here](https://github.com/canonical-ols/javan-rhino/pull/39/files#diff-1fdf421c05c1140f6d71444ea2b27638R22)

katiefenn:
@earnubs: Yep. We had a similar implementation at my last job where authentication was handled on the server and the login status passed to the client in initial state. The landing page was a React component that dispatched an action to set a logged in state on the server-side - so I didn't think this was a million miles away from what we'd end up doing.

Does that make sense? 😅

earnubs:
@kfenn i think so, I'm not 100% clear on the advantage of dispatching an action in that case if you've "hydrated" state from a server rendered initial store state, but let's implement something and see.

(In this case it's not state entering the store per se, it's the initialization of a store with no action on the client to set or update that bit of state, deliberately so.)

katiefenn:
@earnubs: Yes - the login action was called on the server-side so that when the state is hydrated on the client `loggedIn` would already be set to `true`.

It's probably a very good idea to discuss this again once we have an implementation 😉

bartaz:
Looks good to merge. Let's start using it and continue/improve as we need.

PR #43, opened 2016-09-07
fixes #8 make lint failure exit with 0

earnubs:
Preventing npm from capturing the error and throwing ELIFECYCLE error.

Could also do this for test:run, but that might mask other actual NPM errors, so I'll look into that separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "start-prod": "NODE_ENV=production node dist/server/",
     "build": "./scripts/build.sh",
     "watch-client": "node webpack/webpack-dev-server.js",
-    "lint": "eslint -c .eslintrc.js '**/*.js'",
+    "lint": "eslint -c .eslintrc.js '**/*.js'; exit 0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "start-prod": "NODE_ENV=production node dist/server/",
     "build": "./scripts/build.sh",
     "watch-client": "node webpack/webpack-dev-server.js",
-    "lint": "eslint -c .eslintrc.js '**/*.js'",
+    "lint": "eslint -c .eslintrc.js '**/*.js'; exit 0",
```
bartaz: Isn't it needed in other commands too?

Test's failing (`test:run`) also seem to make npm complain.

earnubs: Yeah I think that would mask other npm related errors, since the test scripts are little more complex, I'm going to look at some other solutions there.

bartaz:
Ok, LGTM.

PR #49, opened 2016-09-07
42: Speed up unit tests

katiefenn:
- Removed babel-plugin-webpack-loaders
- Replaced webpack-loaders plugin with some basic require extension stubs
- Split up tests into separate tasks so that quick unit tests can be run in isolation


package.json
```diff
@@ -10,9 +10,11 @@
     "build": "./scripts/build.sh",
     "watch-client": "node webpack/webpack-dev-server.js",
     "lint": "eslint -c .eslintrc.js '**/*.js'; exit 0",
-    "test": "npm run test:run && npm run lint",
-    "test:run": "BABEL_DISABLE_CACHE=1 NODE_ENV=TEST nyc --require babel-core/register mocha --compilers js:babel-register --recursive",
-    "test:watch": "npm run test:run -- --watch",
+    "test": "nyc npm run test:run && npm run lint",
+    "test:run": "npm run test:unit && npm run test:routes",
+    "test:unit": "BABEL_DISABLE_CACHE=1 NODE_ENV=TEST mocha --require babel-core/register --compilers js:babel-register ./test/unit/setup.js ./test/unit/**/*",
```
katiefenn: Split up test tasks so that speedy unit tests can be run on their own.

The unit test task takes about 3s to run on its own.

package.json
```diff
@@ -10,9 +10,11 @@
...
+    "test:routes": "BABEL_DISABLE_CACHE=1 NODE_ENV=TEST mocha --require babel-core/register --compilers js:babel-register ./test/routes/setup.js ./test/routes/**/*",
+    "test:watch": "npm run test:unit -- --watch",
```
katiefenn: Does anyone want to run the routes tests, coverage and linting in watch?

test/unit/setup.js
```diff
@@ -0,0 +1,6 @@
+// Stub out loading of CSS dependencies
+require.extensions['.css'] = () => {};
```
katiefenn: Technically, `require.extensions` is deprecated. The node docs say that they're probably going to be around for a long time, and I'm not too concerned about using them outside of production.

test/unit/components/t_footer.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import expect from 'expect';
 import { shallow } from 'enzyme';
 
-import Footer from '../../src/components/footer';
+import Footer from '../../../src/components/footer';
```
bartaz: I wonder if there is a way to simplify these paths, not to nest so deeply.

I'm not sure about `__dirname`, it will probably still be in test directory rather then project directory, but maybe we could find some other way?

bartaz:
Looks good and runs as fast as it should be 👏

PR #50, opened 2016-09-08
50: Rendering Redux initial state on server

katiefenn:
- Added store creation on server
- Added initial state string to <Html> component
- Added initial state to client rendering

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -1,14 +1,17 @@
 import Express from 'express';
 import MongoConnect from 'connect-mongo';
-import Html from '../src/helpers/html';
 import React from 'react';
+import { createStore } from 'redux';
 import session from 'express-session';
-import conf from './configure.js';
-import login from './routes/login';
-import routes from '../src/routes';
 import util from 'util';
 import { match, RouterContext } from 'react-router';
 import { renderToString } from 'react-dom/server';
+
+import reducers from '../src/redux/';
+import conf from './configure.js';
+import login from './routes/login';
+import routes from '../src/routes';
+import Html from '../src/helpers/html';
 
 const MongoStore = MongoConnect(session);
 const sessionStore = new MongoStore({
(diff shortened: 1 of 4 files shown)


server/server.js
```diff
@@ -1,15 +1,18 @@
...
 import util from 'util';
 import { match, RouterContext } from 'react-router';
 import { renderToString } from 'react-dom/server';
 
+import reducers from '../src/redux/';
+import conf from './configure.js';
+import login from './routes/login';
+import routes from '../src/routes';
+import Html from '../src/helpers/html';
+
```
katiefenn: Separated out first-party dependencies from third party dependencies.

server/server.js
```diff
@@ -41,14 +44,21 @@ function serve(webpackIsomorphicTools) {
       } else if (redirectLocation) {
         res.redirect(302, redirectLocation.pathname + redirectLocation.search);
       } else if (renderProps) {
+        const store = createStore(reducers);
+
```
katiefenn: Store is now created on the server as well as the client. No actions are called at present, but some will probably be added here soon.

src/helpers/html.js
```diff
@@ -23,6 +21,8 @@ export default class Html extends Component {
         </head>
         <body>
           <div id="content" dangerouslySetInnerHTML={{ __html: content }}/>
+          <script
+            dangerouslySetInnerHTML={{ __html: `window.__PRELOADED_STATE__ = ${JSON.stringify(preloadedState)}` }} />
```
katiefenn: This is where the initial state is injected into the page body to be sent to the client.

src/index.js
```diff
@@ -12,7 +12,8 @@ import { Provider } from 'react-redux';
 import routes from './routes';
 import reducers from './redux/';
 
-const store = createStore(reducers);
+const preloadedState = window.__PRELOADED_STATE__;
+const store = createStore(reducers, preloadedState);
```
katiefenn: This is where the initial state is re-hydrated.

PR #52, opened 2016-09-08
48 api routes for snap purchases

earnubs:
express routes for store purchases api, with inclusion of macaroon auth header

### DIFF
--- a/server/handlers/login.js
+++ b/server/handlers/login.js
@@ -13,7 +13,7 @@ export const getMacaroon = (req, res, next) => {
     url: `${conf.get('UBUNTU_SCA:URL')}/dev/api/acl/`,
     method: 'POST',
     json: {
-      'permissions': ['package_access']
+      'permissions': ['package_access', 'package_purchase']
     }
   };
 
@@ -27,7 +27,6 @@ export const getMacaroon = (req, res, next) => {
     req.session.cid = extractCaveatId(body.macaroon);
     next();
   });
-
 };
 
 export const authenticate = (req, res) => {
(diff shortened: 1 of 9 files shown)


server/handlers/login.js
```diff
@@ -13,7 +13,7 @@ export const getMacaroon = (req, res, next) => {
     url: `${conf.get('UBUNTU_SCA:URL')}/dev/api/acl/`,
     method: 'POST',
     json: {
-      'permissions': ['package_access']
+      'permissions': ['package_access', 'package_purchase']
```
earnubs: Really we only need package_purchase today, but when we want to look up details of a snap we'll need package access.

server/middleware/index.js
```diff
@@ -0,0 +1,19 @@
+export const requireAuthentication = (req, res, next) => {
```
earnubs: This will need to be handled with a message to the user that they need to relogin, typically this is a modal window with message and login link.

server/middleware/index.js
```diff
@@ -0,0 +1,19 @@
...
+  if (!req.session.authenticated) {
+    res.status(401).send({
+      'message': 'No session'
+    });
+  }
+
+  next();
+};
+
+export const requireAuthorization = (req, res, next) => {
```
earnubs: Ditto. This will need to be handled with a message to the user that they need to relogin, typically this is a modal window with message and login link.

test/routes/t_api.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    // mock the request to SCA
+    const sca = nock(conf.get('UBUNTU_SCA:URL'))
+      .matchHeader('authorization', authorization)
+      .filteringRequestBody(() => {
+        return body;
+      })
+      .post('/purchases/customers', body)
+      .reply(200);
```
katiefenn: I know that this code is only needed for this test case, but it would be nice to have it in a `beforeEach` block. If only because I can fold it away when opening test files.

earnubs: added #56

Issue #42, opened 2016-09-06
Plain anchor in header nav has no styles


earnubs:
Header navigation link style is coupled to NavLink, but not all navigation links in the header are NavLinks (login, logout).

Possibly the link style should be part of the header component, and the NavLink specify only styles that are specific to that component (which may be none?).

PR #54, opened 2016-09-09
user auth state in redux store

earnubs:
Add user object into store from server side.


earnubs:
Fixes #42

src/reducers/index.js
```diff
@@ -0,0 +1,26 @@
+import * as ActionTypes from '../actions';
+import { combineReducers } from 'redux';
+
+// FIXME do we need this?
```
earnubs: I think yes, because even though we don't add this data to the store via an action, this defines the shape of the store.

src/components/user/root.css
```diff
@@ -0,0 +1,27 @@
+$link-color: $ubuntu-orange;
```
bartaz: @earnubs @katiefenn It may be worth to decide on some CSS module file naming convention.

We use `index.js` for main file (for nicer imports), for now I was using `componentName.css` (like `user.css`) for css file, but maybe we prefer using something like `root.css` here or `index.css`?

I personally prefer component name in CSS name, so that when importing I know what it is: `import styles from './user.css';`.

katiefenn: I agree, our modular stylesheets worked best when we had one for each named component (eg. `user.js` + `user.css`). We sometimes had a common stylesheet for shared styles (`common.css` or `shared.css`).

server/server.js
```diff
@@ -46,7 +45,22 @@ function serve(webpackIsomorphicTools) {
...
+            req.session.teams.length
+        );
+
+        const initialState = {
+          authenticatedUser: {
+            isAuthenticated: req.session.authenticated,
+            isDev: isDev,
+            name: req.session.name
+          }
+        };
```
katiefenn: Can this initial state be moved to a Redux reducer file?

earnubs: The reducer already sets the `authenticatedUser` shape, and how to handle that piece of data, this is setting the values and needs to be here because we're grabbing data from the request.

src/components/user/index.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component, PropTypes } from 'react';
+
+import style from './user.css';
+
+export default class User extends Component {
+  render() {
+    const { user } = this.props;
+    let link;
+    if (user.isAuthenticated) {
+      link = <div className={ style.link }><span className={ style.username }>{ user.name }</span> <a href="/logout">Logout</a></div>;
```
katiefenn: This might be easier to read if split up into multiple lines.

src/components/user/index.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component, PropTypes } from 'react';
+
+import style from './user.css';
+
+export default class User extends Component {
+  render() {
+    const { user } = this.props;
+    let link;
+    if (user.isAuthenticated) {
+      link = <div className={ style.link }>
```
katiefenn: This could be tidied further by moving the JSX out of the render method:

```
renderAuthenticatedLink(user) {
  return (
    ...
  );
}
```

src/components/user/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default class User extends Component {
+  renderAuthenticatedLink(user) {
+    let link;
+
+    if (user.isAuthenticated) {
+      link = <div className={ style.link }>
+        <span className={ style.username }>{ user.name }</span>{ ' ' }
+        <a href="/logout">Logout</a>
+      </div>;
```
katiefenn: This could be tidied further by moving the JSX out of the render method:

```
renderAuthenticatedLink(user) {
  return (
    ...
  );
}
```

bartaz: Yep, JSX is nice until you hit conditionals.

PR #66, opened 2016-09-14
fixes #64

### LINKED ISSUE #64: wrong route for card editing, per spec
should be `/payment/edit` according to spec, which is what it would be on the cli

earnubs:
make url match one in spec, and therefore url in cli


bartaz:
Looks good. I added `/pay/add-card` when prototyping form. Good to have real one in place already.

PR #69, opened 2016-09-14
use cheap-module-eval-source-map for faster rebuild speed in dev

earnubs:
Didn't see a difference from source-maps when an error was thrown, and this is reportedly faster.


katiefenn:
👍

PR #70, opened 2016-09-15
Basic form validation

bartaz:
Added payment form fields state to form component and simple validation (checking for required fields on submit and on blur).

Not using redux-form causes our form component to grow with some boilerplate code (a lot `onChange` and `onBlur` handler passing to all inputs), but at least it's only the form component that is stateful and inputs are just dump render components.

More validation (using Stripe.js for card details checking) will follow in separate branch.

### DIFF
--- a/src/components/forms/form.js
+++ b/src/components/forms/form.js
@@ -1,7 +1,7 @@
 import React, { PropTypes } from 'react';
 
 export default function Form(props) {
-  return <form onSubmit={ props.onSubmit }>
+  return <form onSubmit={ props.onSubmit } noValidate={true}>
     { props.children }
   </form>;
 }
--- a/src/components/forms/inputField.css
+++ b/src/components/forms/inputField.css
@@ -2,6 +2,9 @@ $border-color: $mid-grey;
 $border-color-focus: $warm-grey;
 $background-color: #FFF;
 
+$error-color: $error;
+$success-color: $success;
+
 .inputField {
   margin-bottom: 0.5em;
 }
@@ -42,9 +45,9 @@ $background-color: #FFF;
 }
 
 .textInput.error {
-  border-color: red;
+  border-color: $error-color;
 }
 
 .textInput.success {
-  border-color: green;
+  border-color: $success-color;
 }
--- a/src/components/forms/inputField.css
+++ b/src/components/forms/inputField.css
@@ -40,3 +40,11 @@ $background-color: #FFF;
   border-color: $border-color-focus;
   outline: none;
 }
+
+.textInput.error {
+  border-color: red;
+}
+
+.textInput.success {
+  border-color: green;
+}
--- a/src/components/payments-form/index.js
+++ b/src/components/payments-form/index.js
@@ -7,8 +7,96 @@ import { CheckboxField, Fieldset, FieldRow, Form, InputField, SelectField } from
 import styles from './payments-form.css';
 
 export default class PaymentsForm extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      fields: this.getInitialValues()
+    };
+  }
+
+  getInitialValues() {
+    const initialState = {
+      value: '',
+      valid: false,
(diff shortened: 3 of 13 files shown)


src/components/forms/inputField.css
```diff
@@ -40,3 +40,11 @@ $background-color: #FFF;
...
   outline: none;
 }
+
+.textInput.error {
+  border-color: red;
+}
+
+.textInput.success {
+  border-color: green;
+}
```
bartaz: Will update to colors from vanilla later on.

src/components/payments-form/index.js
```diff
@@ -24,48 +129,172 @@ export default class PaymentsForm extends Component {
...
-        <Fieldset>
           <h4>Name and address</h4>
-          <InputField name="customerFullname" label="Full name" placeholder="John Doe" />
-          <InputField name="customerAddress1" label="Address line 1" placeholder="e.g 20 Ingram Street" />
-          <InputField name="customerAddress2" label="Address line 2" placeholder="Optional" />
-          <InputField name="customerState" label="State/County" placeholder="e.g Essex" />
+          <InputField
+            label="Full name"
+            placeholder="John Doe"
+            {...this.state.fields.customerFullname}
```
bartaz: Passing each field's state and event handlers to every input component... but I guess there is no way around it without decorator such as redux-form.

src/components/forms/form.js
```diff
@@ -1,7 +1,7 @@
 import React, { PropTypes } from 'react';
 
 export default function Form(props) {
-  return <form onSubmit={ props.onSubmit }>
+  return <form onSubmit={ props.onSubmit } noValidate={true}>
```
bartaz: Disabling native browser form validation (for required fields), so ours kicks in.

src/components/payments-form/index.js
```diff
@@ -7,10 +7,115 @@ import { CheckboxField, Fieldset, FieldRow, Form, InputField, SelectField } from
...
+    const fields = {};
+    // initialise all fields with default values and state
+    names.forEach(name => fields[name] = { ...initialState, name });
+    // mark optional field as not required
+    optional.forEach(name => fields[name].required = false);
+
+    return fields;
+  }
+
+  validate(fields, { forceTouched }={}) {
```
katiefenn: Could this be separated out into a generic validation function so it can be tested and re-used?

src/components/payments-form/index.js
```diff
@@ -7,10 +7,115 @@ import { CheckboxField, Fieldset, FieldRow, Form, InputField, SelectField } from
...
+    Object.keys(fields).forEach((name) => {
+      const field = fields[name];
+      // field is required but empty
+      if (field.required && !field.value) {
```
katiefenn: Could this be separated out into a generic validation function so it can be tested and re-used?

src/components/payments-form/index.js
katiefenn: Could this be separated out into a generic validation function so it can be tested and re-used?

src/components/payments-form/index.js
earnubs: Or as a static method?

bartaz: @katiefenn I extracted validating non empty values to separate `validation` module (other validation functions can be added there).
I also added tests for this method in PaymentsForm class tests, to have it covered.

test/unit/components/t_payments_form.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  context('validation', () => {
+    let validate;
+
+    beforeEach(() => {
+      validate = wrapper.instance().validate;
+    });
+
+    it('should be implemented', () => {
```
katiefenn: We can probably get away with not doing tests like these. If `validate` satisfies the tests that check its behaviour, it's safe to assume it's implemented 😉

katiefenn:
LGTM 👍

PR #71, opened 2016-09-16
Card validation

bartaz:
Added card validation with Stripe.js and error messages to fields.
Improved styling of the form.


src/validation/index.js
```diff
@@ -5,3 +5,9 @@ export function validateNonEmpty(value) {
 
   return !!value;
 }
+
+/* global Stripe */
+// use global Stripe from Stripe.js for validation in browser
+export const validateCardNumber = (number) => Stripe.card.validateCardNumber(number);
```
bartaz: We need to keep `Stripe` references inside function calls, because our modules are called before Stripe is loaded, so referencing it during import makes Stripe undefined.

In runtime, when function is called Stripe is already available.

src/components/forms/inputField.js
```diff
@@ -1,37 +1,29 @@
...
+  const className = `${styles.inputField} ${styles[size]}`;
+  const inputClassName = `${styles.textInput} ${styles[status]}`;
+  const labelClassName = `${styles.label} ${styles[status]}`;
+  const errorMsgClassName = `${styles.errorMsg} ${styles[status]}`;
 
   return <div className={ className }>
-    <Label htmlFor={ id }>{ label }:</Label>
-    <input id={ id } name={ name } type={ type }
+    <label htmlFor={ id } className={ labelClassName }>{ label }:</label>
+    <input id={ id } name={ props.secret ? null : name } data-name={ name } type={ type }
```
katiefenn: Having both inline and indented attributes looks a bit untidy. Why not indent them all?

katiefenn: I really think having a dedicated credit card field would be better than having a `secret` prop. It really makes it that much harder to send sensitive information where it shouldn't be.

src/components/forms/selectField.js
```diff
@@ -9,13 +8,15 @@ export default function SelectField(props) {
...
 
+  const className = `${styles.inputField} ${styles[size]}`;
+  const labelClassName = `${styles.label} ${styles[status]}`;
+  const errorMsgClassName = `${styles.errorMsg} ${styles[status]}`;
+
   return <div className={ className }>
-    <Label htmlFor={ id }>{ label }:</Label>
-    <Select id={ id } name={ name }
+    <label htmlFor={ id } className={ labelClassName }>{ label }:</label>
+    <Select id={ id } name={ props.secret ? null : name } data-name={ name }
```
katiefenn: Having both inline and indented attributes looks a bit untidy. Why not indent them all?

src/components/forms/selectField.js
```diff
@@ -9,13 +8,15 @@ export default function SelectField(props) {
   const { name, label, size='full', required, options } = props;
   const id = `ID_SELECT_FIELD_${name}`;
 
-  const className = `${styles.inputField} ${styles[size]}`;
-
   const status = props.touched ? (props.valid ? 'success' : 'error') : null;
 
+  const className = `${styles.inputField} ${styles[size]}`;
```
katiefenn: If all props are indented, there should be enough room to use these values in the JSX instead of assigning them to variables.

src/components/forms/inputField.js
bartaz: @katiefenn Having validation and all form logic in form component rather than inputs makes custom 'credit card' component actually just a text input with name hidden. Then making expiry and CVC components that do exactly the same, just have different name.

How about renaming `secret` prop, to `sensitive` or `isSensitive` in InputFields.

And maybe create `SensitiveInputField` (other name suggestions welcome) that would just be InputField with sensitive prop turned on, that we could use for any sensitive data?

src/validation/index.js
katiefenn: 👍

katiefenn:
LGTM 👍

PR #73, opened 2016-09-19
Billing address

bartaz:
Added binding between customer and billing address when checkbox is clicked.

If billing address is meant to be the same as customer address, billing inputs will be disabled and their values populated from customer address.

Also added tests for input onChange handler and input values binding.

### DIFF
--- a/src/components/forms/inputField.js
+++ b/src/components/forms/inputField.js
@@ -7,7 +7,7 @@ export default function InputField(props) {
   const id = `ID_INPUT_FIELD_${name}`;
   const status = props.touched ? (props.valid ? 'success' : 'error') : null;
 
-  return <div className={ `${styles.inputField} ${styles[size]}` }>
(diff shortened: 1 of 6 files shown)


src/components/forms/inputField.js
```diff
@@ -3,11 +3,11 @@ import React, { PropTypes } from 'react';
 import styles from './inputField.css';
 
 export default function InputField(props) {
-  const { name, label, type='text', size='full', placeholder } = props;
+  const { disabled, name, label, type='text', size='full', placeholder } = props;
   const id = `ID_INPUT_FIELD_${name}`;
   const status = props.touched ? (props.valid ? 'success' : 'error') : null;
 
-  return <div className={ `${styles.inputField} ${styles[size]}` }>
+  return <div className={ `${styles.inputField} ${styles[size]} ${disabled ? styles.disabled : ''}` }>
```
bartaz: Is it still fine as a string, or should we use `classNames` already?

earnubs: Still fine by me.

PR #74, opened 2016-09-20
Client side config

earnubs:
tl;dr Server side config is `settings/index.js` via `import server/configure.js`, and has access to all settings; client side is via `src/config` and has access to only the config in `UNIVERSAL` in `settings/index.js`. Tests are setup with config via `test/setup.js` where we stub out `global` with `UNIVERSAL` from server config.

We want config available on the server and the client, with the ability to set a default in a file (settings/index.js) but override it with environment variables.

Problem: nconf depends on fs, which is a node API, not available in the browser.
Solution: Webpack has a DefinePlugin which allows setting global vars at build time, essentially it does a search an replace on any tokens defined in it's setup, which we can pull from nconf.
Problem: DefinePlugin doesn't understand ES6 modules, so it skips any 'import time' definitions of the tokens it's supposed to be replacing with values from config.

Solution: set globals via initial page render, from the server via nconf

Problem: tests run in node, `window` does not exist there, but `global`
Solution: quietly stub out `global` in node so we don't error and fail to start the server.


.eslintrc.js
```diff
@@ -1,48 +1,50 @@
 module.exports = {
-    'env': {
-        'browser': true,
-        'es6': true,
-        'node': true,
-        'mocha': true
+  'globals': {
```
earnubs: Added a global, fixed indentation (hence diff)

server/configure.js
```diff
@@ -3,8 +3,7 @@ var base = require('../settings');
 
 // allow these overrides from env
 nconf.env([
-  'APP:PORT',
-  'APP:HOST',
+  'APP_URL',
```
earnubs: This is a little less error prone I think.

server/macaroons/index.js
```diff
@@ -1,5 +1,8 @@
 import { MacaroonsBuilder } from 'macaroons.js';
 import conf from '../configure';
+import url from 'url';
+
+const UBUNTU_SSO_HOST = url.parse(conf.get('UBUNTU_SSO_URL')).host;
```
earnubs: I wonder if further down the line (or now?) we might want to do this kind of url parsing and splitting in configure.js, and make URLS a special case where they are split up into components when added via settings.

server/openid/relyingparty.js
```diff
@@ -25,7 +25,7 @@ export default (cid) => {
   }
   return new openid.RelyingParty(
     conf.get('OPENID:VERIFY_URL'),
-    conf.get('OPENID:REALM'),
+    conf.get('APP_URL'),
```
earnubs: No need to have 2 config vars for the same thing, and this explains what is going  on here a little better.

src/actions/customer.js
```diff
@@ -43,7 +43,7 @@ export function postStripeToken(token) {
   return (dispatch) => {
     dispatch(sendStripeToken(token));
 
-    return fetch('http://localhost:3000/api/purchases/customers', {
+    return fetch(`${APP_URL}/api/purchases/customers`, {
```
earnubs: This is a global, injected via DefinePlugin

earnubs:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;5k9OesD6&#x2F;65-config-in-browser-and-server"> Config in browser and server</a>

src/actions/customer.js
```diff
@@ -43,7 +45,9 @@ export function postStripeToken(token) {
   return (dispatch) => {
     dispatch(sendStripeToken(token));
 
-    return fetch('http://localhost:3000/api/purchases/customers', {
+    const url = conf.get('MU_URL');
```
bartaz: Just a small detail, we can do it later, but it would be nice to have same config item names both on server and client.

Currently it's `conf.get('UNIVERSAL:MU_URL')` on server and `conf.get('MU_URL')` on client. I would be nice to have either `MU_URL` or `UNIVERSAL:MU_URL` on both.

test/unit/actions/t_customer.js
```diff
@@ -3,11 +3,12 @@ import configureMockStore from 'redux-mock-store';
...
 
+import conf from '../../../server/configure.js';
 import {
   postStripeToken,
-  sendStripeToken,
-  sendStripeTokenSuccess,
-  sendStripeTokenFailure
+    sendStripeToken,
+    sendStripeTokenSuccess,
+    sendStripeTokenFailure
```
earnubs: editor failure :)

PR #80, opened 2016-09-20
Stripe token actions

bartaz:
Added actions/reducers to send card data to Stripe and store token.

Card data is now collected and sent to Stripe (only billing address is used).
Token is received and stored in Redux.

### DIFF
--- a/src/actions/stripe.js
+++ b/src/actions/stripe.js
@@ -1,0 +1,55 @@
+export const CREATE_STRIPE_TOKEN = 'CREATE_STRIPE_TOKEN';
+export const CREATE_STRIPE_TOKEN_SUCCESS = 'CREATE_STRIPE_TOKEN_SUCCESS';
+export const CREATE_STRIPE_TOKEN_FAILURE = 'CREATE_STRIPE_TOKEN_FAILURE';
+
+export function createStripeToken(cardData) {
+  return {
+    type: CREATE_STRIPE_TOKEN,
+    cardData
+  };
+}
+
+export function createStripeTokenSuccess(token, response) {
+  return {
+    type: CREATE_STRIPE_TOKEN_SUCCESS,
+    token,
+    response
+  };
+}
+
+export function createStripeTokenFailure(error) {
+  return {
+    type: CREATE_STRIPE_TOKEN_FAILURE,
+    error
+  };
+}
+
+/* global Stripe */
+export function postCardData(cardData) {
+  return (dispatch) => {
+
+    dispatch(createStripeToken(cardData));
+
+    Stripe.setPublishableKey('pk_test_6pRNASCoBOKtIshFeQd4XMUh');
+
+    Stripe.card.createToken({
+      number: cardData.cardNumber,
+      cvc: cardData.securityNumber,
+      exp: cardData.expiryDate,
+
+      name: cardData.name,
+      address_line1: cardData.addressLine1,
+      address_line2: cardData.addressLine2,
+      address_city: cardData.city,
+      address_state: cardData.state,
+      address_zip: cardData.postcode,
+      address_country: cardData.country
+    }, (status, response) => {
+      if (response.error) {
+        dispatch(createStripeTokenFailure(response.error));
+      } else {
+        dispatch(createStripeTokenSuccess(response.id, response));
+      }
+    });
+  };
+}
--- a/src/actions/stripe.js
+++ b/src/actions/stripe.js
@@ -3,6 +3,8 @@ export const SAVE_VALIDATED_CARD_DATA = 'SAVE_VALIDATED_CARD_DATA';
 export const CREATE_STRIPE_TOKEN_SUCCESS = 'CREATE_STRIPE_TOKEN_SUCCESS';
 export const CREATE_STRIPE_TOKEN_FAILURE = 'CREATE_STRIPE_TOKEN_FAILURE';
 
+import { postStripeToken } from './customer';
+
 export function createStripeToken(formCardData) {
   return {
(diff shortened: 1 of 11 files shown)


src/actions/stripe.js
```diff
@@ -0,0 +1,55 @@
...
+  };
+}
+
+/* global Stripe */
+export function postCardData(cardData) {
+  return (dispatch) => {
+
+    dispatch(createStripeToken(cardData));
+
+    Stripe.setPublishableKey('pk_test_6pRNASCoBOKtIshFeQd4XMUh');
```
bartaz: Key needs to be taken from configuration.

Also, this could be set only once when Stripe is initialised, but I couldn't find a better place to do it ATM. It seems that loading stripe.js with Helmet makes it load async in client, and having static script that initialises stripe resulted is Stripe being undefined yet.

src/actions/stripe.js
```diff
@@ -0,0 +1,55 @@
...
+      address_line2: cardData.addressLine2,
+      address_city: cardData.city,
+      address_state: cardData.state,
+      address_zip: cardData.postcode,
+      address_country: cardData.country
+    }, (status, response) => {
+      if (response.error) {
+        dispatch(createStripeTokenFailure(response.error));
+      } else {
+        dispatch(createStripeTokenSuccess(response.id, response));
```
bartaz: Should it also dispatch the `postStripeToken` action to send token to pay immediately? Or should it be handled elsewhere (where?)?

src/actions/stripe.js
```diff
@@ -0,0 +1,55 @@
...
+
+/* global Stripe */
+export function postCardData(cardData) {
+  return (dispatch) => {
+
+    dispatch(createStripeToken(cardData));
+
+    Stripe.setPublishableKey('pk_test_6pRNASCoBOKtIshFeQd4XMUh');
+
+    Stripe.card.createToken({
```
bartaz: Again, having Stripe as global is a pain for testing. Any simple way to mock globals for tests?

src/actions/stripe.js
earnubs: These are the bits we'll refine over time. I would say that very possibly there is no `createStripeTokenSuccess` action, only a `sendStripeToken` action https://github.com/canonical-ols/javan-rhino/blob/master/src/actions/customer.js#L7 dispatched in response to a successful promise resolution in the 'createStripeToken' action.

And the same for failure, that it's a general failure action with a payload from this error, dispatched via a TBC `flashUserMessage` action creator.

src/reducers/stripe.js
```diff
@@ -0,0 +1,33 @@
...
+        ...state,
+        isFetching: true,
+        cardData: action.cardData
+      };
+    case ActionTypes.CREATE_STRIPE_TOKEN_SUCCESS:
+      return {
+        ...state,
+        isFetching: false,
+        token: action.token,
+        response: action.response
```
earnubs: a little vague, but then i think this action doesn't exist :)

src/actions/stripe.js
earnubs: On failure, we might need some code to decide what to do, but I think the options are "try again" or "tell the user something went wrong"

test/unit/reducers/t_stripe.js
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    it('should add token to store', () => {
+      expect(stripe({}, action)).toInclude({ token: 'test_token_0987654321' });
+    });
+
+    it('should add stripe response object to store', () => {
+      expect(stripe({}, action)).toInclude({
+        response: {
+          id: 'test_token_0987654321',
```
earnubs: This test makes me wonder if there are 2 actions:

1) sendStripeToken
2) saveCardValidatedData (that is, Stripe has validated it as well as the client)

earnubs: why would we save that card data, Stephen?

bartaz: @earnubs I would save card data received from Stripe to make it possible to show success message to user saying something like "Your card **** **** 4242 for John Doe was successfully saved, please go back to CLI and use snap buy command again to make a purchase".

But I agree we can have 2 separate actions for that.

src/actions/stripe.js
bartaz: @earnubs Agreed.

src/actions/stripe.js
bartaz: Let's move it to separate branch once clientside config is done.

src/actions/stripe.js
bartaz: I'm keeping `createStripeTokenSuccess` and `createStripeTokenSuccess` actions for now, as they handle Stripe `isFetching` status.

src/reducers/stripe.js
earnubs: To clarify my 'vague' comment, I think that placing the whole response in is a little vague, I can't look at the reducer and tell what the store should look like.

src/actions/stripe.js
katiefenn: I've used Sinon's sandboxing feature: http://sinonjs.org/docs/#sandbox

src/actions/stripe.js
katiefenn: In [the example here](https://stripe.com/docs/api#authentication), the `api_key` option appears to allow the API key to be passed with each call to the API.

src/reducers/stripe.js
bartaz: @earnubs I removed setting whole response in the store in favour of just setting validatedCardData that we get from Stripe.

But I forgot to update initial state here, so need to fix that.

PR #85, opened 2016-09-22
update openid to latest, move back to node 6

earnubs:
Abandon the node 4 ship.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "4"
+  - "6"
   - "node"
 services:
   - mongodb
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "4"
+  - "6"
```
earnubs: These 2 lines are essentially the same, so I'm happy to remove the "node" line if we think it's totally redundant and a waste of time today, but I somewhat like the idea of the tests rolling up to the next version without us having to remember.

earnubs:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;HLajwOhy&#x2F;57-use-node-6"> use node 6</a>

PR #91, opened 2016-09-23
remove universal conf block

earnubs:
Updates based on review of #74

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -66,7 +66,7 @@ function serve(webpackIsomorphicTools) {
 
         // config we share from server side to client side
         const config = {
-          MU_URL: conf.get('MU_URL')
+          UNIVERSAL: conf.get('UNIVERSAL')
         };
 
         // You can also check renderProps.components or renderProps.routes for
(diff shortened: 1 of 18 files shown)


test/unit/actions/t_customer.js
```diff
@@ -57,7 +57,7 @@ describe('async actions', () => {
   let scope;
 
   before(() => {
-    GLOBAL.MU_URL = conf.get('MU_URL');
+    global.MU_URL = conf.get('MU_URL');
```
earnubs: "GLOBAL deprecated, use global" warning

server/server.js
```diff
@@ -64,6 +64,11 @@ function serve(webpackIsomorphicTools) {
         };
         const store = configureStore(initialState);
 
+        // config we share from server side to client side
```
earnubs: You still need/want to import those settings into nconf, to allow a setting to be overridden from env vars, etc.

The only simple way I can think to do this, without ending up with 2 nconf instances, is make settings look like: 

```
{
  UNIVERSAL: {
  ...
  },
  SERVER: {
  ...
  }
}

 // nconf.get('UNIVERSAL:MU_URL');

```

But to keep symmetry in get signature the universal settings would need to be rendered into HTML as:

```
__CONFIG__ = {
  UNIVERSAL: conf.get('UNIVERSAL')
}

// conf.get('UNIVERSAL:MU_URL');
```

Which is close to what you're asking for I think?

earnubs: PS. in this PR, `settings.js` still defines everything, you just have to choose what is released to the browser explicitly in the render method of root (/). The proposal above is to make what is released to the browser implicit in the inclusion of it in `UNIVERSAL`

katiefenn: 👍

PR #93, opened 2016-09-23
Simple thank you message after successful API response, mocked API

bartaz:
Simple success message that we can show up after successful response.
I guess we will need to confirm final copy text for that.

Also mocked Pay API for now, so we can test successful responses not relying on SCA for now.

screen shot 2016-09-23 at 15 20 07

### DIFF
--- a/server/handlers/api.js
+++ b/server/handlers/api.js
@@ -6,18 +6,36 @@ const UBUNTU_SCA_URL = conf.get('UBUNTU_SCA_URL');
 export const customers = (req, res) => {
   const auth = req.session.authorization;
 
-  const options = {
-    uri: `${UBUNTU_SCA_URL}/purchases/v1/customers`,
-    headers: {
-      authorization: auth
-    },
-    json: true,
-    body: {
-      stripe_token: req.body.stripe_token
-    }
-  };
+  // FIXME:
+  // send real requests when we don't need mocked ones anymore
+  // const options = {
+  //   uri: `${UBUNTU_SCA_URL}/purchases/v1/customers`,
+  //   headers: {
+  //     authorization: auth
+  //   },
+  //   json: true,
+  //   body: {
+  //     stripe_token: req.body.stripe_token
+  //   }
+  // };
 
-  request.post(options).pipe(res);
+  // request.post(options).pipe(res);
+
+  // FIXME:
+  // sending mocked responses until
+  if (auth) {
+    res.status(200).json({
+      'accepted_tos_date': '2016-09-13T11:27:31+00:00',
+      'latest_tos_date': '2016-09-09T12:00:00+00:00',
+      'latest_tos_accepted': true,
+      'has_payment_method': true
+    });
+  } else {
+    res.status(401).json({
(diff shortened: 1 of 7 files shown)


server/handlers/api.js
```diff
@@ -6,18 +6,36 @@ const UBUNTU_SCA_URL = conf.get('UBUNTU_SCA_URL');
...
+  //   body: {
+  //     stripe_token: req.body.stripe_token
+  //   }
+  // };
 
-  request.post(options).pipe(res);
+  // request.post(options).pipe(res);
+
+  // FIXME:
+  // sending mocked responses until
```
bartaz: Mocking API for now.

src/actions/customer.js
```diff
@@ -60,7 +60,7 @@ export function postStripeToken(token) {
     })
       .then(checkStatus)
       .then(response => response.json())
-      .then(json => dispatch(sendStripeTokenSuccess(json.tos_accepted)))
+      .then(json => dispatch(sendStripeTokenSuccess(json.latest_tos_accepted)))
```
bartaz: It seems it's supposed to be `latest_tos_accepted` according to API spec: https://myapps.developer.staging.ubuntu.com/docs/api/snap-purchases.html#operation/createCustomer

src/components/customer-success/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+export function CustomerSuccess(props) {
+  if (!props.customer.tosAccepted) {
+    return null;
+  }
+
+  return (
+    <div className={ styles.customerSuccess }>
+      <h3>Thank you</h3>
+      <p>Your payment details were successfully saved. Proceed with your purchases in command line using <code className={ styles.code }>snap buy</code> command.</p>
```
earnubs: Make the success clear in the header, so a user familiar with the process can skip the next line.

```
<h3>Payment details successfully saved, thank you!</h3>
<p>Please proceed with your purchase on the command line using ...</p>
```

src/containers/add-card.js
```diff
@@ -19,6 +21,7 @@ export class AddCard extends Component {
       <div className={ styles.container }>
         <PaymentsForm />
         <PaymentDetails />
+        <CustomerSuccess />
```
earnubs: My understanding would be that the container does the work and the components try to be as dumb as possible, but we can refactor post GA.

test/routes/t_api.js
```diff
@@ -43,18 +43,20 @@ describe('purchases api', () => {
       'stripe_token': 'foo'
     };
 
+    // FIXME: responses are currently mocked in API
```
earnubs: Why remove the mock? This is a test of the route handler, not SCA's response.

earnubs: Silly me, you're not calling SCA any more ... (I was thinking you could have nocked SCA rather than fake the response, that's where I got confused)

src/components/customer-success/index.js
earnubs: Approving anyway, because I don't have final say on the copy.

PR #97, opened 2016-09-24
export countries so as not to trigger warning

### LINKED ISSUE #96: File [~/Workspace/javan-rhino/src/components/payments-form/countries.js] ignored, nothing could be mapped
Warning on `countries.js` when running tests, it's not exporting correctly I think.

earnubs:
The export style is correct I think (based on http://www.2ality.com/2014/09/es6-modules-final.html#default-exports-%28one-per-module%29) but it triggers a warning, fixes #96


katiefenn:
LGTM

PR #104, opened 2016-09-27
Welcome user, and ask user to login or show user details

earnubs:


src/components/button/index.js
```diff
@@ -2,15 +2,28 @@ import React, { PropTypes } from 'react';
...
+  type: PropTypes.string,
+  appearance: React.PropTypes.oneOf(['primary', 'secondary'])
+};
+
 export default function Button(props) {
   const { appearance='primary', ...rest } = props;
   return <button {...rest} className={ style[appearance] } />;
 }
 
+export function Anchor(props) {
```
katiefenn: Can we put this in a new file called `anchor.js`? I'm fairly sure I would have to resort to using grep to find it if it's in a file called `button.js`.

src/components/payments-form/payments-form.css
```diff
@@ -1,4 +1,18 @@
...
+  content: '';
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  opacity: 0.8;
+  background-color: #fff;
+  display: block;
+  z-index: 2147483647;
```
katiefenn: Why the mahoosive z-index?

src/components/button/index.js
earnubs: This is a button, it just happens to be a anchor disguised as a button ... I wouldn't expect `anchor.js` to give me a button.

src/components/payments-form/payments-form.css
earnubs: That's x browser maximum z-index, to ensure nothing ever sits above this.

katiefenn: @earnubs: I guessed - but why do such elements need to be always-on-top?

src/components/sign-in-banner/index.js
```diff
@@ -0,0 +1,40 @@
...
+    );
+  }
+
+  render() {
+    const { identity } = this.props;
+
+    return (
+      <div className={ styles.banner }>
+        { identity.isAuthenticated ? this._renderAuthenticated(identity)
+        : this._renderUnauthenticated() }
```
katiefenn: Picky: Might be nice to align the latter two operands:

```
{ identity.isAuthenticated
    ? this._renderAuthenticated(identity)
    : this._renderUnauthenticated() }
```

katiefenn: Suggestion: is it worth defining renderAuthenticated as a function instead of a method, making it properly private?

earnubs:
[Ask user to log in before they try to submit the form](https://trello.com/c/Ermwa1Eq/73-ask-user-to-log-in-before-they-try-to-submit-the-form)

src/components/welcome/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import styles from './welcome.css';
+
+export default function Welcome()  {
+  return (
+    <div className={ `${styles.welcome} ${styles.box}` }>
```
katiefenn: Suggestion: Use `<article>` instead of `<div>`.

test/unit/components/t_payments_form.js
```diff
@@ -24,6 +27,29 @@ describe('<PaymentsForm /> component', () => {
     expect(wrapper.find(Form).length).toBe(1);
   });
 
+  it('renders with disabled class', () => {
```
katiefenn: We can probably get away without doing this. It's possible that we could change the CSS for disabled buttons and this test could start reporting a false positive.

src/components/button/index.js
katiefenn: It could be `button/anchor.js` though, right? I guess what I'm cheering for is one-component-per-file, really.

earnubs: It could be, but I would prefer to keep components like this (stateless) in a single file, and we can even tell eslint:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md#ignorestateless

src/components/payments-form/payments-form.css
earnubs: This 'modal' is to  disable all the controls within the container, and match the design; I could have picked 1000, or 10000 but with this z-index I'm making clear this should be on top.

src/components/welcome/index.js
earnubs: If you can tell me one advantage to `<article>`, maybe :)

test/unit/components/t_payments_form.js
earnubs: Not sure I understand?

src/components/button/index.js
katiefenn: 👍

src/components/payments-form/payments-form.css
katiefenn: 👍 Okay, cool.

Optional: Maybe it's worth making it a variable to make it extra-readable?

src/components/welcome/index.js
katiefenn: @earnubs: It's useful for assistive technologies and screen readers? http://webaim.org/blog/future-web-accessibility-html5-semantic-tags/

I appreciate accessibility is not a high priority at the moment, but it would at least allow people to read the content on the home page a bit easier.

test/unit/components/t_payments_form.js
katiefenn: I don't think there's any value in testing CSS classes, because it's not the classes themselves that provide any value. It's the styles that matter. Testing classes is a false economy.

src/components/payments-form/payments-form.css
earnubs: Yep, should have done that, will do.

src/components/welcome/index.js
earnubs: It's only implemented in a couple of screen readers afaik, and it's usefulness is debatable in that context; I don't believe "Article: Welcome Katie" is better than "Welcome Katie".

test/unit/components/t_payments_form.js
earnubs: removed

src/components/sign-in-banner/index.js
earnubs: done

test/unit/components/t_payments_form.js
```diff
@@ -24,6 +27,25 @@ describe('<PaymentsForm /> component', () => {
...
+          isAuthenticated: true
+        },
+        stripe: {},
+        customer: {},
+        dispatch: () => {}
+      };
+
+      wrapper = shallow(<PaymentsForm {...props} />);
+    });
+    it('renders without disabled class', () => {
```
katiefenn: One more test to be removed.

PR #107, opened 2016-09-28
Added test for macaroon error handling

bartaz:
Added test for macaroon error handling, fixed crashing server on missing macaroon.


server/handlers/login.js
```diff
@@ -20,14 +20,13 @@ export const getMacaroon = (req, res, next) => {
...
   request(options, (error, response, body) => {
-    // TODO handle macaroon failure
-    if (error) {
-      res.send('Get macaroon failed');
+    if (error || !body.macaroon) {
+      res.status(401).send({ error: 'Get macaroon failed' });
+    } else {
+      req.session.macaroon = body.macaroon;
+      req.session.cid = extractCaveatId(body.macaroon);
+      next();
```
bartaz: @earnubs Not sure if `next()` should be called only on success, or always? In case of calling it on errors should we pass errors to `next()`?

server/server.js
```diff
@@ -57,6 +57,7 @@ function serve(webpackIsomorphicTools) {
 
         const initialState = {
           identity: {
+            errors: req.session.errors,
```
bartaz: Should we keep errors in `identity` (as rest of data from session) or create another reducer just to handle flash/error messages from session?

server/handlers/login.js
bartaz: Handler was updated to be a redirect, and `next` needs to be called only on success - otherwise express complains on headers being set multiple times on response (I guess it tries to redirect more then once).

server/server.js
katiefenn: +1 new reducer.

katiefenn: "errors" could be a property of standard shape on one or more of the reducers, allowing each to manage their own errors. I think this would be useful, because we won't necessarily want to report all errors in the same way.

bartaz:
Fixed by #111

PR #110, opened 2016-09-28
52: Notifications

katiefenn:
- Added new Notification component
- Added new actions and reducers for notifications
- Added success notification to the successful "add card" scenario
- Added error notification to the failure "add card" scenario
- Added new "hasActionOfType" assertion to help match Redux actions


src/style/variables.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   'dark-aubergine':      '#2c001e', // dark aubergine (enterprise)
   'mid-grey':            '#cdcdcd',
   'warm-grey':           '#888888', // warm grey
+  'dark-grey':           '#666666', // dark grey
```
katiefenn: @earnubs Agreed - especially when it's the difference between `#888888` and `#666666`. I'll raise this with barrymcgee today.

bartaz:
I think it would be better to show notification under the header, not on top of it - as it looks quite weird that we cover heading.

Also, full width notification looks quite strange on wide screens (mine is quite extreme, but it shows the point):

screen shot 2016-09-29 at 13 33 05

So maybe we should limit notification width to container with.

katiefenn:
@bartaz: The problem with locating notifications relative to the header is that the header will often appear off-screen, especially when filling out longer forms. Locating the notifications relative to the top of the viewport ensures they are visible even when looking at the bottom of the page.

bartaz:
@katiefenn Agreed. I wonder how it would look if we position notification relative to viewport, but with top offset big enough for header. So even if header is off the screen, notification will be visible (but with some offset from the top).

But let's do whatever works for now.

src/components/notification/index.js
```diff
@@ -0,0 +1,3 @@
+import Notification from './notification';
```
katiefenn: I know this unnecessary, but it'll make the actual component file easier to find with quick-load dialogs. I think we should do this with all single-component modules later.

src/components/notification/notification.css
```diff
@@ -0,0 +1,67 @@
...
+  background-size: 16px 16px;
+  margin: 0;
+  text-align: left;
+  padding-left: 1.5rem;
+}
+
+.response-success {
+  /* TODO: Extract image */
+  composes: response;
+  background-image: url("data:image/svg+xml,%3Csvg width='17px' height='17px' viewBox='0 0 17 17' version='1.1' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E%3Cg id='Page-1' stroke='none' stroke-width='1' fill='none' fill-rule='evenodd'%3E%3Cg id='notification-success' transform='translate(1.000000, 1.000000)'%3E%3Cg id='Page-3---colours'%3E%3Cg id='Notifications---single'%3E%3Cg id='Group'%3E%3Cg id='ICON'%3E%3Ccircle id='circle6710' stroke='$success' stroke-width='1.5' fill='$success' cx='7.2500086' cy='7.2500086' r='7.2500086'%3E%3C/circle%3E%3Cpolygon id='path6712' fill='#ffffff' points='11.0502986 4.1734486 10.9843986 4.2311486 6.2496486 8.3783686 3.4740786 5.9974286 2.6350186 6.9463086 6.2503386 10.7500186 11.7500086 4.9627786 11.0502986 4.1734886'%3E%3C/polygon%3E%3C/g%3E%3C/g%3E%3C/g%3E%3C/g%3E%3C/g%3E%3C/g%3E%3C/svg%3E");
```
katiefenn: Is it worth extracting these images to a separate file, or is that better achieved in a new branch?

test/unit/actions/t_customer.js
```diff
@@ -74,14 +88,11 @@ describe('async actions', () => {
...
 
-    const expectedActions = [
-      { type: ActionTypes.SEND_STRIPE_TOKEN, token },
-      { type: ActionTypes.SEND_STRIPE_TOKEN_SUCCESS, tosAccepted }
-    ];
-
     return store.dispatch(postStripeToken(token))
       .then(() => {
-        expect(store.getActions()).toEqual(expectedActions);
+        expect(store.getActions()).toHaveActionOfType(
```
katiefenn: I've added a custom assertion (toHaveActionOfType) to make asserting an action has taken place easier.

test/unit/setup.js
```diff
@@ -9,3 +11,18 @@ const conf = require('../../server/configure.js');
 global.__CONFIG__ = {
   UNIVERSAL: conf.get('UNIVERSAL')
 };
+
+// Custom assertions
+expect.extend({
```
katiefenn: The toHaveActionOfType assertion basically determines if an array of action objects (from getActions()) contains an object with a type property matching a given string.

webpack/dev-config.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
   module: {
     loaders: require('./loaders-config.js')
   },
-  devtool: '#cheap-module-eval-source-map',
+  devtool: 'source-map',
```
katiefenn: Think this might have already been done in another branch.

src/components/notification/notification.css
earnubs: Do you mean a separate file of data urls or load them from disk like normal people?

test/unit/setup.js
earnubs: I would consider putting this in a helpers module and importing it, so it's more obvious / less magic.

webpack/dev-config.js
earnubs: Doesn't hurt to do it twice.

src/components/notification/notification.css
katiefenn: @earnubs: Load from disk.

test/unit/setup.js
katiefenn: Cool, can do that in a different branch. 👍

katiefenn: @earnubs: Assertion moved to a new file.

src/components/notification/index.js
earnubs: ..but mess up tab completion on `notification.js` and `notification.css` (which I think should be `root.css` but that's for another day)

src/components/notification/notification.css
```diff
@@ -0,0 +1,67 @@
+.notification {
+  /* TODO: Extract image */
+  background-color: $light-grey;
+  border: 0;
+  border-color: $warm-grey;
+  border-radius: .125rem;
```
earnubs: use a pixel value rather than calculate from an em unless the value you want is directly related to the font (like the space between paragraphs), radii are in proportion to other radii or are set from a visual style

PR #111, opened 2016-09-29
Error handling in login route

earnubs:
Add basic error handling middleware for login, put error on session and display error with an Oyez, a top posted, page level notification pattern.


server/handlers/login.js
```diff
@@ -20,9 +17,12 @@ export const getMacaroon = (req, res, next) => {
   };
 
   request(options, (error, response, body) => {
-    // TODO handle macaroon failure
     if (error) {
-      res.send('Get macaroon failed');
+      return next(error);
```
earnubs: "If you pass anything to the next() function (except the string 'route'), Express regards the current request as being in error and will skip any remaining non-error handling routing and middleware functions."

server/handlers/login.js
```diff
@@ -60,17 +58,23 @@ export const verify = (req, res) => {
...
     }
     res.redirect('/');
   });
 };
+
+export const errorHandler = (err, req, res, next) => {
+  if (req.session) {
+    req.session.error = err;
+  }
+  res.redirect('/');
```
earnubs: Render the error into initial state, to be picked up by @katiefenn notifications #110

bartaz: @earnubs We assume there can be only one error in session? Also - we want it to be instance or Error? If we want to pass it to client redux store we need it to be serializable object.

server/handlers/login.js
bartaz: When I was testing/mocking error responses I had a case that when returning 500 from mocked API didn't result with `error` still being `null` in this handler, while `body.macaroon` was obviously undefined. That's why I check for body.macaroon in #107 https://github.com/canonical-ols/javan-rhino/pull/107/files#diff-c8d78287f5e38fb0b1eb58acb553f29dR23

server/handlers/login.js
earnubs: @bartaz yeah, this is pretty naive about errors in session. I guess I need to append to an errors list and have the notifications handler clean that up when they are dismissed @katiefenn ?

server/handlers/login.js
earnubs: @bartaz thanks, i didn't think that would be possible :(

earnubs:
[Error handling during login](https://trello.com/c/STcu38ln/88-error-handling-during-login)

PR #113, opened 2016-09-29
Allow log path, host and port to be configured from nconf

1stvamp:
Also fixes nconf env usage a little so that setting `FOO__BAR` and doing `config.get('FOO:BAR')` will actually work.


earnubs:
[Make access log location configurable](https://trello.com/c/2f0uIMwb/83-make-access-log-location-configurable)

PR #115, opened 2016-09-30
Add a juju charm

1stvamp:
This change adds an embedded juju charm under the `charm/` directory, and a Makefile to invoke building the charm (which embeds the production `dist/` build of javan-rhino into it's payload), with working memcached relation to setup the session store.

I also modified how the `MEMCACHE_SESSION_HOST` is set in the session handler, by `.split(',')`-ing it into an array of hosts, so we could also modify the variable into `HOSTS` to be more explicit, but left that for another change.

FYI - some of the OLS charm layer components are not OSS, but that will change in the future I don't foresee anyone outside Canonical wanting to build a juju charm before GA. :smiling_imp: 

https://trello.com/c/EuqWRv9K


Makefile
```diff
@@ -0,0 +1,75 @@
...
+	tar cz --exclude-vcs --exclude-from build-tar-exclude.txt -f $(PAYLOAD) .
+
+## build the charm and payload
+build: $(PAYLOAD)
+
+deploy: build
+	juju deploy local:$(CHARM_SERIES)/$(NAME)
+	juju deploy memcached
+	juju add-relation $(NAME) memcached
+	juju set $(NAME) session_secret='its a secret' deployment=$(DEPLOY_ENV)
```
earnubs: Just checking you mean to have a session secret here.

charm/layer.yaml
```diff
@@ -0,0 +1,18 @@
...
+includes:
+    - layer:ols
+    - interface:memcache
+options:
+    ols:
+        service_name: javan-rhino
+        user: ubuntu
+    basic:
+        packages:
+            - nodejs
```
earnubs: Need node v6

charm/config.yaml
```diff
@@ -0,0 +1,11 @@
+options:
+    deployment:
```
bloodearnest: layer-ols provides this, as 'environment', so you don't need to add it.

Makefile
1stvamp: The deploy target is just for checking the charm works, so yeah this is just a dummy value to make sure it works.

charm/config.yaml
1stvamp: Thanks, fixed.

charm/layer.yaml
1stvamp: As mentioned on IRC, this is tricky, so we'll fix it in the mojo spec for now and as soon as there's a public PPA we can add to the charm's "apt" config we will do.

earnubs:

PR #117, opened 2016-09-30
Terms of Service

bartaz:
Added TOS page with dummy text and basic styles.

Added TOS acceptance checkbox to the form and blocking submit until they are accepted.


src/components/payments-form/index.js
```diff
@@ -76,7 +78,7 @@ export class PaymentsForm extends Component {
   /* RENDER */
 
   render() {
-    const isFetching = this.props.customer.isFetching || this.props.stripe.isFetching;
+    const isFormReady = this.state.isTosAccepted && !this.props.customer.isFetching && !this.props.stripe.isFetching;
```
earnubs:
```
const isFormReady = this.props.customer.isFetching || this.props.stripe.isFetching;
const isTermsAccepted = this.state.isTosAccepted;

const hasPositiveWavesMoriarty = isFormReady && isTermsAccepted; // https://www.youtube.com/watch?v=Xyh-JpWdGmQ
```

PR #118, opened 2016-10-03
Added static notification under page contents

bartaz:
Error notifications are now static in page content (under the form).

Removed success notifications when card is added (as we show thank you message inline already).

Improved API error text, to suggest users to use different card (as any card issue from stripe results in failed request from API).

screen shot 2016-10-03 at 16 15 12

### DIFF
--- a/src/actions/customer.js
+++ b/src/actions/customer.js
@@ -30,7 +30,7 @@ export function sendStripeTokenFailure(errors) {
     dispatch(showNotification({
       message: [
         'There\'s been a problem adding your card.',
-        'Please verify that card details you provided are correct or use a differenct card and try again.'
+        'Please verify that card details you provided are correct or use a different card and try again.'
       ].join(' '),
       status: 'error',
       action: () => {
--- a/src/actions/customer.js
+++ b/src/actions/customer.js
@@ -18,11 +18,6 @@ export function sendStripeToken(token) {
 
 export function sendStripeTokenSuccess(tosAccepted) {
   return (dispatch) => {
-    dispatch(showNotification({
-      message: 'Your card has been added successfully',
-      status: 'success'
-    }));
-
     dispatch({
       type: SEND_STRIPE_TOKEN_SUCCESS,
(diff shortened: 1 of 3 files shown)


bartaz:
At this point it looks quite weird to have them both (fixed on top and under the content).
Also, when there is success it seems that notification is not needed at all - we already show success 'thank you' message, so having notification (especially again on top and bottom) seems too much.

Should we only show bottom notification when form is submitted?

katiefenn:
@bartaz @earnubs: I think part of the problem is that we don't have a good UI pattern for in-line form errors. Anything we do is going to look weird to some degree.

earnubs:
I think inline form errors are fine (in that there is a definition for them in Vanilla under p-notification) it's with "flash messages" we have nothing in place, and using p-notification there just does not work (wrong height, wrong colour, wrong margins, is really a singleton in the UI ... )

So let's deprecate notifications for now, and go with in page errors.

src/actions/customer.js
```diff
@@ -35,7 +30,7 @@ export function sendStripeTokenFailure(errors) {
     dispatch(showNotification({
       message: [
         'There\'s been a problem adding your card.',
-        'Please wait a bit and try again soon.'
+        'Please verify that card details you provided are correct or use a differenct card and try again.'
```
katiefenn: Should we be showing the user a specific error message when this failure can be caused by any number of things, including network problems?

vilagithub: s/differenct card/different card/

Randomly spotted, haven't read the rest ;)

bartaz: @katiefenn Right now API in case of any error returns exactly the same response, so on our side, we cannot tell what was the cause of the error.

Network issues - of course, they may happen, but I guess it's much more likely that there may be some error with a card or payment. So we need to at least suggest to the user that this may be the issue.

bartaz: @vilagithub Well spotted!

PR #121, opened 2016-10-04
set cookie maxAge so it expires before macaroon

earnubs:
also set cookie to `secure` in production and staging


earnubs:
[shorten session length to less than macaroon life](https://trello.com/c/abYoXmcJ/94-shorten-session-length-to-less-than-macaroon-life)

earnubs:
[Express best practice](https://trello.com/c/X0gsN8qI/41-express-best-practice)

PR #122, opened 2016-10-04
drop redux logger

earnubs:
post GA put this back if we're in development, but not staging or production


src/store/configureStore.js
```diff
@@ -9,10 +8,8 @@ export default function configureStore(preloadedState) {
     preloadedState,
     compose(
       applyMiddleware(
-        thunk,
-        createLogger()
-      ),
-      typeof window === 'object' && typeof window.devToolsExtension !== 'undefined' ? window.devToolsExtension() : f => f
```
katiefenn: Why is the Redux devtool extension not being bootstrapped any more?

Issue #116, opened 2016-09-30
Missing name/email address from SSO


bartaz:
If user doesn't share his name and/or email address from SSO UI has missing info:

<img width="948" alt="screen shot 2016-09-30 at 13 45 43" src="https://cloud.githubusercontent.com/assets/83575/18990749/4e74d2f0-8714-11e6-8c63-09cb2d705e6c.png">

Not sure if it's just staging SSO issue, or can we make name (and email?) required?
If we don't want to require those, UI should not expect them and adapt text.

earnubs:
In `server/openid/relyingparty.js`: 

```
new openid.SimpleRegistration({
  'email' : 'required',
  'fullname' : 'required'
}),
```

earnubs:
MU needs to become a trusted RP in SSO, in order to properly support `required` sreg attrs, which I'll talk to @canonical-ols/ols about today

bartaz:
Fixed by #125 #130

bartaz:
Also thanks to trusted RP email and name are properly required now.

PR #125, opened 2016-10-04
Making user details from SSO optional

bartaz:
Related to #116

Making name and email optional in welcome message and header.


src/components/sign-in-banner/index.js
```diff
@@ -10,11 +10,15 @@ export default class SignInBanner extends Component {
 
     function renderAuthenticated(identity) {
       const { name, email } = identity;
+
+      const welcome = name ? `Welcome, ${name}` : 'Welcome';
```
earnubs: default value for name? "Friend"?

PR #127, opened 2016-10-04
update default cookie name, per express sec best practice

earnubs:
https://expressjs.com/en/advanced/best-practice-security.html#dont-use-the-default-session-cookie-name


earnubs:
[Express best practice](https://trello.com/c/X0gsN8qI/41-express-best-practice)

PR #128, opened 2016-10-04
add mu to title

earnubs:
daring.


src/containers/app.js
```diff
@@ -17,7 +17,7 @@ export class App extends Component {
         <Helmet
           htmlAttributes={{ 'lang': 'en' }}
           title='buy [snap-name]'
-          titleTemplate='my.ubuntu.com - %s'
+          titleTemplate='μ | my.ubuntu.com | %s'
```
katiefenn: I can see this change causing friction.

Geddit?

PR #129, opened 2016-10-04
include script blocks in server render

earnubs:
avoid throwing error on validation because stripe loaded after bundle


earnubs:
This causes Stripe to load twice, because of this bug https://github.com/nfl/react-helmet/issues/149

Note: it has nothing to do with Stipe channel.html also loading the script.

src/helpers/html.js
```diff
@@ -17,9 +17,16 @@ export default class Html extends Component {
           {head.title.toComponent()}
           {head.meta.toComponent()}
           {head.link.toComponent()}
+          {head.script.toComponent()}
           <link rel="stylesheet" href={ assets.styles.main } />
         </head>
         <body>
+          //  https://github.com/nfl/react-helmet/issues/149
```
earnubs: We could just simply add via <script > below, but I wanted to keep a record of this

PR #132, opened 2016-10-04
Allow NodeJS 6 to be installed by charm from vendored .deb file

1stvamp:
This allows Node to be installed from a .deb in files/nodejs.deb, so we can get shiny Node 6.7 on prod servers.
It also fixes some of the build targets in the Makefile which got a little skewiff in the git-build PR.

Note: line 47 onwards of `javan_rhino.py` is the relevant section that changed, which isn't immediately obvious due to typoed filename change (`java_rhino.py` :arrow_right: `javan_rhino.py`).


1stvamp:

PR #133, opened 2016-10-04
Fix 'charm build' error on repeated git-build invocations

1stvamp:
`charm build` will complain when building into a non-empty charm build dir that includes any files not listed on a layer, which is what happens when we build into our builds git repo.

To overcome this I've added a `EXTRA_CHARM_BUILD_ARGS` param to the Makefile and then set this to `--force` for the `git-build` target, so that on this target `charm build` willy merrily continue regardless of any extra files in the target repo.


1stvamp:

vilagithub:
seems a bit brutal but makes sense

PR #134, opened 2016-10-04
Use rsync to update build git repo

1stvamp:
By using `rsync` to synchronise built charm files we can ensure that a bunch of VCS ditritus and build files (`.done`, `.build.manifest` for example) are excluded from build asset git repo when invoking `make git-build`.


1stvamp:

PR #137, opened 2016-10-05
Moved static notification to top and scrolling when it shows up

bartaz:


src/containers/app.js
```diff
@@ -11,6 +11,14 @@ import styles from './app.css';
 export class App extends Component {
   render() {
     const { identity, oyez, notifications } = this.props;
+
+    if (notifications.notification) {
+      setTimeout(() => {
+        console.log("scrollTo")
```
earnubs: i expect lint to shout at you

PR #138, opened 2016-10-05
Add missing wsgi and postgres charm interfaces

1stvamp:
Although unused by javan-rhino charm, they are required by the base OLS charm layer.


earnubs:

PR #144, opened 2016-10-05
Fix bind host/port for staging

1stvamp:
Should have set this same as production when I did the charm, oops.


1stvamp:

Issue #88, opened 2016-09-22
ubuntu logo is inline / data url


earnubs:
Wouldn't it be better served and cached?

earnubs:
Fixed in url loader file length limit.

PR #145, opened 2016-10-05
88: Remove inlined logo

katiefenn:
- Extracted main logo into separate file by tweaking WebPack loader config

### DIFF
--- a/webpack/loaders-config.js
+++ b/webpack/loaders-config.js
@@ -28,7 +28,7 @@ module.exports = [
   },
   {
     test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
-    loader: 'url-loader?limit=10000&mimetype=image/svg+xml'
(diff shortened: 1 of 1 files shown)


webpack/loaders-config.js
```diff
@@ -28,7 +28,7 @@ module.exports = [
   },
   {
     test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
-    loader: 'url-loader?limit=10000&mimetype=image/svg+xml'
+    loader: 'url-loader?limit=4000&mimetype=image/svg+xml'
```
katiefenn: Main logo is a little over 4k in size, so it will not be inlined with the new settings.
I decided not to set this to a much lower value, because I thought some things might be nice to bundle with the stylesheet: an icon set, for example.

katiefenn:
Fixes #88

earnubs:
Superceded by #186, which drops url-loader

PR #146, opened 2016-10-05
Fix the paths used for rsync

1stvamp:
The rsync destination was incorrect, it was transfering a subdirectory instead of the whole tree.

This also adds a conditional to cloning the build repo.


1stvamp:

Issue #154, opened 2016-10-12
Clicking dropdown icon in country field does not open country list


thomir:
When adding payment details, the country section drop-down has a little triangle image in the right hand side to indicate that this is a drop-down field. However, clicking the triangle does not open the field - at least in Chrome on Ubuntu version 53.0.2785.116. Clicking on the main part of the widget opens the drop-down, but this behaves differently to every other standard drop-down widget, and is somewhat confusing.

PR #159, opened 2016-10-14
154: Fix select element arrow icon

katiefenn:


katiefenn:
As a separate DOM element, the arrow icon does not trigger a "click" event on the select input when it is clicked.

This has been fixed by positioning the select input in front of the arrow icon using a higher z-index.

katiefenn:
Fixes #154

PR #162, opened 2016-10-18
Switch to raw payloads for charm builds

1stvamp:
Currently we have very big deployment payload deltas due to tarballing the build tree (and git not handling binary deltas), this uses the latest changes in the OLS base layer to enable vendoring the raw build tree without using a tarball, so we have much smaller payload deltas.


1stvamp:
@bloodearnest can you review when you get a min, please?

Issue #170, opened 2016-10-20
"problem adding your card" notification persists too long


katiefenn:
The "problem adding your card" notification persists even when the user logs out. This is because it is of the sort that requires user acknowledgement. Preferably, the notification should either auto-dismiss after a few seconds, or dismiss when the user navigates away from the "page" showing the error.

<img width="974" alt="screen shot 2016-10-20 at 14 46 56" src="https://cloud.githubusercontent.com/assets/1410204/19560391/a3feceb8-96d4-11e6-8244-689358f2fc27.png">

PR #172, opened 2016-10-20
170: Card error notification persists too long

katiefenn:
- Changed card error notification to auto-dismiss instead of requiring user action.


katiefenn:
Fixes #170

Issue #99, opened 2016-09-26
Components do not update in server-side rendering when changes saved to source


katiefenn:
Steps to recreate:
1) Start development site
2) Make changes to a component and save changes
3) Observe components hot-reloading
4) Refresh the page in the browser
5) Observe the "flash" between the old component look and the new component look as the app is rendered.

bartaz:
Ad 5) There is always also a warning in browser console:

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
```

earnubs:
fixed in #185

PR #185, opened 2016-11-02
99: Hot reloading server

katiefenn:
- Server.js now watches src directory for changes and will reload components when this happens.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "babel-preset-stage-0": "^6.16.0",
     "babel-register": "^6.16.3",
     "body-parser": "^1.15.2",
+    "chokidar": "^1.6.1",
     "classnames": "^2.2.5",
     "connect-memcached": "^0.2.0",
     "coveralls": "^2.11.14",
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -4,10 +4,12 @@ import helmet from 'helmet';
 import morgan from 'morgan';
 import path from 'path';
 import session from 'express-session';
+import chokidar from 'chokidar';
 
 import * as routes from './routes/';
 import conf from './configure';
 import sessionConfig from './helpers/session';
+import { clearRequireCache } from './helpers/hot-load';
 
 const logsPath = conf.get('SERVER:LOGS_PATH') || path.join(__dirname, '../../logs/');
 const accessLogStream = fs.createWriteStream(
@@ -31,5 +33,18 @@ app.use(Express.static(__dirname + '/../public', { maxAge: '365d' }));
 app.use('/', routes.login);
 app.use('/api', routes.api);
 app.use('/', routes.universal);
+
+if (process.env.NODE_ENV === 'development') {
+  // Do "hot-reloading" of express stuff on the server
+  // Throw away cached modules and re-require next time
+  // Ensure there's no important state in there!
+  const watcher = chokidar.watch('./src');
+
+  watcher.on('ready', function() {
+    watcher.on('all', function() {
+      clearRequireCache(/[\/\\]src[\/\\]/, require.cache);
+    });
+  });
+}
 
 export { app as default };
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2678,7 +2678,7 @@
       "version": "1.0.3"
     },
     "object.omit": {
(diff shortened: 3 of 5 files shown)


katiefenn:
Tested on Yakkety with npm 3.10.7.

While the specific problem of hot-reloading server components is fixed, I found that watched files would only trigger watch events on Ubuntu every so often. I estimate this to be every 60 seconds or maybe on the fifth or sixth write to a file. This issue also affects react-hot-loader, so I suspect this could be an issue with [chokidar](https://github.com/paulmillr/chokidar).

src/server/server.js
```diff
@@ -9,13 +9,13 @@ import url from 'url';
 import util from 'util';
 import { match, RouterContext } from 'react-router';
 import { renderToString } from 'react-dom/server';
+import chokidar from 'chokidar';
```
katiefenn: This is the same dependency used by `babel-cli`.

src/server/server.js
```diff
@@ -41,6 +41,7 @@ app.use('/api', api);
 function serve(webpackIsomorphicTools) {
   app.use(Express.static(__dirname + '/../public'));
   app.use('/', function (req, res) {
+    let routes = require('../common/routes').default;
```
katiefenn: Reload routes to make sure the latest version of components are used. In practice, this does not happen on every request due to node's require cache.

The cache is later emptied if changes are detected.

src/server/server.js
```diff
@@ -119,4 +120,21 @@ function serve(webpackIsomorphicTools) {
...
+
+  watcher.on('ready', function() {
+    watcher.on('all', function() {
+      Object.keys(require.cache).forEach(function(id) {
+        if (/[\/\\]src[\/\\]common[\/\\]/.test(id)) {
+          delete require.cache[id];
+        }
+      });
+    });
+  });
```
katiefenn: This clears the require cache if changes to components are detected.

Follows an example found [here](https://github.com/glenjamin/ultimate-hot-reloading-example/blob/master/server.js#L33).

katiefenn:
Fixes #99

katiefenn:
Copied from ##frontend:

<kfenn> "On Mac OS X, chokidar by default uses a native extension exposing the Darwin FSEvents API. This provides very efficient recursive watching compared with implementations like kqueue available on most *nix platforms. Chokidar still does have to do some work to normalize the events received that way as well."
18:04 Hmm!
18:05 I wonder if limiting the scope of file watching can help its performance.
18:05 
<roadmr> apparently on Ubuntu it has to use that kqueue thing or inotify, which need to start one watcher per subdirectory
18:06 so if you manage to prune down the set of dirs under watch, that would certainly help perf
18:06 
<kfenn> Katie Fenn Cool, I will try that tomorrow. Thanks roadmr

src/server/server.js
bartaz: As this inline require is only needed on dev server for hot reload maybe we can keep static import on top and "re-require" it inline only if NODE_ENV === 'development'?
This will be more explicit that it's dev only thing.

I guess a comment that it's for hot loading would be good too, soon we will forget why routes are loaded there.

src/server/server.js
bartaz: Manually removing items from require cache feels dirty... But I guess there is no way around that if we want hot reloading.

src/server/server.js
katiefenn: I can do this, but using only node's `require` function. Variables imported from ES6 modules are read-only.

bartaz: @katiefenn Right... imports can't be overridden in runtime.

Anyway, it would be good to at least be explicit with a comment why the routes are loaded differently. Apart from that it looks good.

package.json
```diff
@@ -43,6 +43,7 @@
     "babel-preset-stage-0": "^6.16.0",
     "babel-register": "^6.16.3",
     "body-parser": "^1.15.2",
+    "chokidar": "https://registry.npmjs.org/chokidar/-/chokidar-1.6.1.tgz",
```
earnubs: Why the url?

bartaz: Url also appears in shrinkwrap because of that.

katiefenn: *blinks*

PR #186, opened 2016-11-02
Css modules require hook

earnubs:
replace webpack isomorphic tools with require hooks for a little less magic, and allow easier refactoring and testing of the server code.


test/routes/t_api.js
```diff
@@ -1,94 +1,133 @@
...
-import { agent } from 'supertest';
+import supertest, { agent } from 'supertest';
+import session from 'express-session';
+
+import api from '../../src/server/routes/api';
+import conf from '../../src/server/configure';
+import sessionConfig from '../../src/server/helpers/session';
 
+/**
+ * Universal route captures all, meaning we can't modify the app route stack
```
earnubs: These tests have some value now in the absence of acceptance tests, but may be considered too fragile once we're testing e2e.

src/server/index.js
```diff
@@ -1,10 +1,21 @@
+const url = require('url');
+const util = require('util');
 require('babel-register');
+require('css-modules-require-hook/preset');
 
-const WebPackIsomorphicTools = require('webpack-isomorphic-tools');
-const basePath = require('path').resolve(__dirname, '../..');
+const conf = require('./configure.js');
+const WEBPACK_DEV_URL = conf.get('SERVER:WEBPACK_DEV_URL') || '';
+require('images-require-hook')('.svg', `${WEBPACK_DEV_URL}/static/icons`);
```
bartaz: So currently with this line we only have support for importing `.svg` images?
What does the `static/icons` path stands for? Do we have such directory with images? I only recall some svg images that are in components folders along with the code.

earnubs: As configured, `.svg` only, but adding support for other types is jus a matter of adding the extension to the first require hook argument, and to the file loader test in webpack config.

`static/icons` is where the images require hook will put the files it encounters on build https://github.com/ptshih/images-require-hook#parameters

earnubs:
[Refactor express server](https://trello.com/c/chxSMtVe/108-refactor-express-server)

PR #188, opened 2016-11-04
Correctly read NODE_ENV from proccess.env

bartaz:
It seems it slipped through the review. I found it when trying to debug something else running prod locally.

### DIFF
--- a/src/server/helpers/html.js
+++ b/src/server/helpers/html.js
@@ -5,7 +5,7 @@ import { Provider } from 'react-redux';
 
 import conf from '../../server/configure';
 
-const assetHost = (global.NODE_ENV === 'production') ? ''
+const assetHost = (process.env.NODE_ENV === 'production') ? ''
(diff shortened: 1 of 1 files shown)


src/server/helpers/html.js
```diff
@@ -5,7 +5,7 @@ import { Provider } from 'react-redux';
 
 import conf from '../../server/configure';
 
-const assetHost = (global.NODE_ENV === 'production') ? ''
+const assetHost = (process.env.NODE_ENV === 'production') ? ''
```
katiefenn: I think this is fine. Storing derived values in config bloats the number of settings, potentially making things harder to understand and maintain.

earnubs: I had a quick look at getting this dead code eliminated, and along with discussion in the channel we'd want to look in the future at DefinePlugin to potentially optimise this more in ugligy, but as it is, it's called, set and cached on module load and shouldn't be a perf bottleneck.

PR #190, opened 2016-11-07
Static resources cache

bartaz:
MaxAge set to 365d, webpack generates resources with hashes in names, using assets plugin to create JSON with assets mapping.


src/server/handlers/universal.js
```diff
@@ -55,13 +55,17 @@ export const handleMatch = (req, res, error, redirectLocation, renderProps) => {
     // below, if you're using a catch-all route.
     const component = <RouterContext {...renderProps} />;
 
+    // TODO: load/import it elsewhere?
+    const assets = require('../../../webpack-assets.json');
```
katiefenn: Given:
- the development site doesn't make use of cachebusting URLs
- we know what the names of the bundle files are, is it worth changing
- we've just made a reasonable effort to ditch `webpack-assets.json` by switching away from Webpack Isomorphic Tools

...is it worth changing this so that webpack-assets.json is only used in production?

earnubs: Not sure if this helps but take a look at https://github.com/canonical-ols/javan-rhino/pull/191/files#diff-a6b43a8832abdd9cd1bf7c30da41d462 where I'm starting the app server from the webpack-dev server success callback, ensuring that webpack is done (I think, I haven't tested that ... the server and the build might not be the same thing!)

src/server/helpers/html.js
```diff
@@ -25,7 +20,7 @@ export default class Html extends Component {
           {head.meta.toComponent()}
           {head.link.toComponent()}
           {head.script.toComponent()}
-          <link rel="stylesheet" href={ `${assetHost}/static/style.css` } />
+          <link rel="stylesheet" href={ assets.main.css } />
```
earnubs: This looks much better than the manifest plugin!

src/server/handlers/universal.js
bartaz: @katiefenn Having only prod import `webpack-assets.json` could still make dev server fail on startup when JSON is missing (first build on clean project).
Also having dev bundle names 'hardcoded' still requires us to differentiate between dev and prod asset host (webpack vs static).

These are all things we can workaround, but it builds up a bunch of code around that, while the best thing would be to have dev and prod behave in same way.

bartaz: @earnubs As discussed in IRC. Starting server after webpack server doesn't help (as build is still not ready), but `waitUntilValid` callback can be used to do that.

test/routes/t_login.js
```diff
@@ -1,4 +1,25 @@
...
+
+// stubbing webpack-assets.json imported by handlers/universal
+const webpackAssetsStub = {
+  main: { js: {}, css: {} },
+  // tell proxyquire not to try to load stubbed file
+  '@noCallThru': true,
+  // tell proxyquire this dependency can be loaded by nested modules
+  // WARNING: this disables require cache, which may cause side effects
+  // see: https://github.com/thlorenz/proxyquire#caveat
+  '@global': true
```
bartaz: That's a bit dirty solution.
We proxyquire server module, but we stub webpack-assets.json that is imported way down the dependency tree. For proxyquire to allow that it needs this `@global` flag to make stub available for all required modules.
But there is a caveat - require cache needs to be disabled to make it happen and this can have unexpected side-effects (if modules are required more than once). So generally using `@global` is discouraged.
See: https://github.com/thlorenz/proxyquire#caveat

bartaz: Fixed by stubbing routes instead of webpack-assets.json.

test/routes/t_login.js
```diff
@@ -1,4 +1,25 @@
-import app from '../../src/server/server.js';
+// load server/server.js via proxyquire to stub webpack-assets.json
+import proxyquire from 'proxyquire';
+
+// import login routes for testing
+import loginRoutes from '../../src/server/routes/login.js';
```
bartaz: Only importing login routes and subbing rest (api and universal) to avoid loading `universal` handler that imports `webpack-assets.json`.

It makes quite a good sense in this test, as it's meant to only test login routes and doesn't need anything else.

PR #192, opened 2016-11-10
Recreated shrinkwrap on ubuntu with npm 3.10.7

bartaz:


bartaz:
Shouldn't be needed as we have #194

PR #194, opened 2016-11-10
npm-fix

katiefenn:
- Migrate webpack and babel packages that depend on fsevents to
  devDependencies, removing them from the shrinkwrap. This intends to
  fix platform-specific dependencies such as fsevents that cause
  cross-platform issues.


package.json
```diff
@@ -34,9 +34,7 @@
   "dependencies": {
     "assets-webpack-plugin": "^3.5.0",
     "autoprefixer": "^6.5.1",
-    "babel-cli": "^6.16.0",
```
katiefenn: I have purposefully limited the packages I have moved to the ones that depend on fsevents for the time-being. I suggest we expand this to all development dependencies if successful.

PR #198, opened 2016-11-15
Added raven middleware to report errors to sentry

bartaz:


src/server/server.js
```diff
@@ -19,6 +20,8 @@ const accessLogStream = fs.createWriteStream(
 
 const app = Express();
 
+app.use(raven.middleware.express.requestHandler(process.env.SENTRY_DSN));
```
katiefenn: Is it worth consuming this setting using nconf (namespaced to only be available on server)? This would allow us to pass in values using env files and CLI switches on the development website.

PR #191, opened 2016-11-07
Talisker format node logging

earnubs:
logs

Replace util.log with Winston configured to output talisker format logs from named logger instances, or the default winston logger.

Log levels have been set to only 'info' and 'debug' (see https://dave.cheney.net/2015/11/05/lets-talk-about-logging). 

Also, express request logging is via winston too, and uses the same talisker format.

Example:
```
2016-11-17 17:19:32.387Z INFO session "Starting with development default session secret"
2016-11-17 17:19:32.394Z INFO express "Express server listening on http://127.0.0.1:3000"
2016-11-17 17:19:33.914Z INFO express "HTTP GET /" statusCode=302, url=/, host=localhost:3000, connection=keep-alive, cache-control=max-age=0, upgrade-insecure-requests=1, user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36", accept=text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8, dnt=1, accept-encoding="gzip, deflate, sdch, br", accept-language=en-GB,en;q=0.8,en-US;q=0.6,fr;q=0.4,fr-FR;q=0.2, method=GET, httpVersion=1.1, originalUrl=/, , responseTime=36
```

To use:

```
import logging from './logging';
const logger = logging.getLogger('myLogger'); // 'myLogger' is the id of this logger instance, and will appear in the log output after log level
logger.info('hello');
```

Will produce:

```
2016-11-08 21:16:16.873Z INFO myLogger "hello"
```

For future use, this should be published as it's own module (along with some other talisker features) `node-talisker`.

PR #201, opened 2016-11-16
Reporting to Sentry when latest_tos_accepted is false

### LINKED ISSUE #184: Sentry: log if `latest_tos_accepted` is false
Once we have sentry set up we should log cases when API returns `latest_tos_accepted` as false after sending card details.

See #151

bartaz:
Fixes #184:


src/common/actions/customer.js
```diff
@@ -17,11 +17,6 @@ export function sendStripeToken(token) {
 }
 
 export function sendStripeTokenSuccess(tosAccepted) {
-  // TODO:
```
bartaz: At first I wanted to add raven call to sentry here in customer action when we actually store the value of `tosAccepted`, but it's run on client in browser where we don't have sentry set up yet (and while our sentry endpoint it's behind the VPN it won't be sent anyway).

Therefore raven call was moved to server side api handler.

src/server/handlers/api.js
```diff
@@ -17,6 +20,14 @@ export const customers = (req, res) => {
     }
   };
 
-  request.post(options).pipe(res);
+  request.post(options, (err, res, body) => {
+    if (err) {
+      client.captureError(err);
```
earnubs: when #191 lands, we also want to log to info the error

PR #202, opened 2016-11-16
117: Configuration overhaul

katiefenn:
- Removed JSON configuration variables
- Moved common settings to env files
- Added missing settings to make-env script
- Removed environment variable whitelist
- Added sentry DSN to Charm config


charm/config.yaml
```diff
@@ -5,3 +5,6 @@ options:
     memcache_session_secret:
         type: string
         description: Secret key for encrypting session data stored in memcached
+    server__sentry_dsn:
+        type: string
+        description: The Sentry DSN that will be used by Raven to send error reports to Sentry
```
katiefenn: New setting for the Sentry DSN.

Ping @1stvamp.

charm/templates/javan-rhino_systemd.j2
```diff
@@ -10,7 +10,6 @@ Environment='SESSION_SECRET={{ session_secret }}'
 Environment='SERVER__LOGS_PATH={{ logs_path }}'
 Environment='SESSION_MEMCACHED_HOST={{ cache_hosts | join(",") }}'
 Environment='SESSION_MEMCACHED_SECRET={{ memcache_session_secret }}'
-Environment='SERVER__PORT=8000'
```
katiefenn: This setting has been moved to env files.

env/development.env
```diff
@@ -0,0 +1,11 @@
...
+SERVER__PORT=3000
+UNIVERSAL__MU_URL=http://localhost:3000
+SERVER__WEBPACK_DEV_URL=http://localhost:3001
+NODE_ENV=development
+DEPLOY_ENV=development
+SERVER__UBUNTU_SSO_URL=https://login.staging.ubuntu.com
+SERVER__UBUNTU_SCA_URL=https://myapps.developer.staging.ubuntu.com
+SERVER__OPENID__VERIFY_URL=http://localhost:3000/login/verify
+SERVER__OPENID__TEAMS="["ubuntuone-hackers"]"
+UNIVERSAL__STRIPE_PUBLISHABLE_KEY=pk_test_daBepdMharNP0PTQYoyQJPjH
```
katiefenn: Common settings moved to environment variables instead. The intention is to remove the hierarchy of configuration in deployed environments as much as possible, so that we are not stuck troubleshooting which settings are taking precedence when debugging problems.

scripts/make-env.js
```diff
@@ -24,6 +24,16 @@ var schema = {
...
+    UNIVERSAL__MU_URL: {
+      description: 'Please enter a base URL for site',
+      default: 'http://localhost:3000'
+    },
+
+    SERVER__WEBPACK_DEV_URL: {
+      description: 'Please enter a WebPack dev server URL',
+      default: 'http://localhost:3001'
+    },
+
```
katiefenn: I've added some missing settings to the make-env script, so that it is more complete.

Many of these settings (and their defaults) are needed so that default answers for all settings produces the `development.env` file needed for starting the development site.

src/server/configure.js
```diff
@@ -1,36 +1,13 @@
 const nconf = require('nconf');
-const development = require('../../settings/development');
-const staging = require('../../settings/staging');
-const production = require('../../settings/production');
 
 nconf.argv();
 
 if (nconf.get('env')) {
-  require('dotenv').config({ path: nconf.get('ENV') });
+  require('dotenv').config({ path: nconf.get('env') });
```
katiefenn: Unrelated bug that stops env files from being loaded properly.

src/server/configure.js
```diff
@@ -1,36 +1,13 @@
...
-    'SERVER__HOST', // SERVER:HOST
-    'SERVER__PORT', // SERVER:PORT
-    'SERVER__LOGS_PATH', // SERVER:LOGS_PATH
-    'SERVER__SENTRY_DSN' // SERVER:SENTRY_DSN
-  ]
+  separator: '__'
 });
 
-if (process.env.NODE_ENV === 'production') {
-  if (process.env.DEPLOY_ENV === 'production') {
```
katiefenn: I think this might be the only place `DEPLOY_ENV` is used. Anyone in favour of going the extra mile and removing it entirely?

test/routes/setup.js
```diff
@@ -4,3 +4,6 @@ require.extensions['.svg'] = () => 'example.svg';
 require.extensions['.png'] = () => 'example.png';
 require.extensions['.gif'] = () => 'example.gif';
 require.extensions['.jpg', '.jpeg'] = () => 'example.jpg';
+
+// Load environment config
+require('dotenv').config({ path: require('path').resolve(__dirname, '../../env/development.env') });
```
katiefenn: This is required now that configure.js isn't doing it.

src/server/configure.js
```diff
@@ -1,36 +1,13 @@
 const nconf = require('nconf');
-const development = require('../../settings/development');
-const staging = require('../../settings/staging');
-const production = require('../../settings/production');
 
 nconf.argv();
 
 if (nconf.get('env')) {
```
bartaz: Could we default to 'env/development.env' here if no `--env` is provided?

This would allow us to still simply run `npm start` to run dev server instead of `npm start -- --env=env/development.env`.

src/server/configure.js
bartaz: 👍 Yes, it's quite confusing to have both on them.

But how would we differentiate between prod and staging without it?

src/server/configure.js
katiefenn: Good idea.

src/server/configure.js
katiefenn: That's the question - do we need to? The point of staging is to have a production build on a non-production platform. It is not mentioned anywhere in our source anymore.

charm/config.yaml
```diff
@@ -5,3 +5,6 @@ options:
     memcache_session_secret:
         type: string
         description: Secret key for encrypting session data stored in memcached
+    server__sentry_dsn:
```
1stvamp: Let's drop the `server__` prefix here to match other charm options, e.g. `sentry_dsn`.

src/server/configure.js
bartaz: @katiefenn @earnubs Yes, current use on .env files (as introduced in this PR) is sufficient for loading different config on production and staging, so we don't need DEPLOY_ENV anymore.

Makefile
```diff
@@ -25,7 +25,6 @@ INTERFACE_PATH = $(TMPDIR)/interface
 CHARM_WHEELDIR = $(TMPDIR)/wheels
 CHARM_DEPS = $(LAYER_PATH)/.done $(INTERFACE_PATH)/.done
 EXTRA_CHARM_BUILD_ARGS ?=
-DEPLOY_ENV ?= devel
```
katiefenn: Ping @1stvamp: I've removed the `DEPLOY_ENV` environment variable.

charm/templates/javan-rhino_systemd.j2
1stvamp: We need something like:
```
Environment='SERVER__SENTRY_DSN={{ sentry_dsn }}'
```

Here, and in `charm/reactive/javan_rhino.py` we need to pass it in from the charm config.

charm/templates/javan-rhino_systemd.j2
```diff
@@ -10,10 +10,8 @@ Environment='SESSION_SECRET={{ session_secret }}'
...
 WorkingDirectory={{ working_dir }}
 User={{ user }}
-Environment='DEPLOY_ENV={{ environment }}'
 ExecStart=/usr/bin/npm run start-build
```
1stvamp: Need to add `--env env/{{ environment }}.env` here I think.

charm/config.yaml
katiefenn: Paging @bartaz...

bartaz: @katiefenn @1stvamp ok, using `SENTRY_DSN` is suggested in sentry docs as well.

This will require small update where current `SERVER:SENTRY_DSN` is loaded - and if we drop `SERVER` prefix we need to make sure we don't leak server side sentry DSN with secret to client.

PR #204, opened 2016-11-16
203: React update breaking build

katiefenn:
- Updated react & react DOM


bartaz:
Already fixed by #206

PR #212, opened 2016-11-18
Talisker status endpoints

earnubs:
See http://talisker.readthedocs.io/en/latest/contract.html#endpoints


src/server/routes/status.js
```diff
@@ -0,0 +1,41 @@
...
+    memcached.end();
+  });
+};
+
+export const error = () => {
+  throw new Error('This is a servicestatus error test');
+};
+
+export const metric = (req, res) => {
+  res.send(200);
```
bartaz: It's just a placeholder for now? How does it differ from `ping`?

earnubs: The route has middleware on it which will send a metric to statsd, the handler is just sending a default OK; would a comment here help?

test/routes/t_login.js
```diff
@@ -5,6 +5,7 @@ import proxyquire from 'proxyquire';
 import loginRoutes from '../../src/server/routes/login.js';
 
 const routesStub = {
+  status: () => {},
```
bartaz: Not sure if it is relevant here (maybe we need whole server), but in other tests we found a way to test a single route without creating an instance of whole server and empty routes stub: https://github.com/canonical-ols/javan-rhino/blob/master/test/routes/t_api.js#L14-L18

May be worth updating now (or at some point).

earnubs: I tried this very quickly and it didn't work, so I'll put it in another PR later.

PR #213, opened 2016-11-18
Set x-vcs-revision header

earnubs:


src/server/middleware/set-revision-header.js
```diff
@@ -0,0 +1,18 @@
...
+// location and name of file is consistent across projects
+const filename = './version-info.txt';
+let revision = 'unknown';
+
+try {
+  // git rev-parse HEAD > version-info.txt
+  revision = fs.readFileSync(filename, 'utf8').trim();
+  // logger.debug(`Got X-VCS-revision as ${revision}`);
+} catch(e) {
+  // logger.info(`missing ${filename} cannot set correct X-VCS-revision)`);
```
earnubs: we don't yet have a logger

src/server/middleware/set-revision-header.js
```diff
@@ -0,0 +1,18 @@
+import fs from 'fs';
+
+// location and name of file is consistent across projects
+const filename = './version-info.txt';
```
katiefenn: Is `version-info.txt` in version control? Should it be?

earnubs: Good question. File is written by the Makefile as part of build, so by default it's not there for us in dev and the header value is 'unknown'. 

So, I think no, it shouldn't be under version control as it's a build artefact.

PR #216, opened 2016-11-21
117: Config overhaul

katiefenn:
- Added path resolution to env file loading
- Re-arranged loading of arguments and env file

### DIFF
--- a/src/server/configure.js
+++ b/src/server/configure.js
@@ -1,15 +1,16 @@
 const path = require('path');
 const nconf = require('nconf');
+const dotenv = require('dotenv');
 
 nconf.argv();
 
 let envFile = 'environments/development.env';
 
 if (nconf.get('env')) {
-  envFile = path.resolve(__dirname, '../..', nconf.get('env'));
+  envFile = path.resolve('./', nconf.get('env'));
 }
 
-require('dotenv').config({ path: envFile });
+dotenv.config({ path: envFile });
 
 nconf.env({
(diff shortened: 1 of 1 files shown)


src/server/configure.js
```diff
@@ -1,12 +1,14 @@
...
 let envFile = 'environments/development.env';
+
 if (nconf.get('env')) {
-  envFile = nconf.get('env');
+  envFile = path.resolve(__dirname, '../..', nconf.get('env'));
 }
 
-nconf.argv();
-
 require('dotenv').config({ path: envFile });
```
earnubs: Maybe move the require to the top?

```
const dotenv = require('dotenv');
...
dotenv.config({ path: envFile });
```

src/server/configure.js
```diff
@@ -1,12 +1,14 @@
+const path = require('path');
 const nconf = require('nconf');
 
+nconf.argv();
+
 let envFile = 'environments/development.env';
+
 if (nconf.get('env')) {
-  envFile = nconf.get('env');
+  envFile = path.resolve(__dirname, '../..', nconf.get('env'));
```
earnubs: Would ./ be easier than __dirname?

katiefenn: path.resolve with __dirname will work if we start the application using npm scripts, `node src/server` or `node dist/server`. It's also quite [a common pattern](http://stackoverflow.com/a/7352912).

PR #218, opened 2016-11-21
Updated react-hot-reloader to 3.0.0-beta.6

bartaz:


webpack/loaders-config.js
```diff
@@ -4,7 +4,7 @@ module.exports = [
   {
     test: /\.js$/i,
     exclude: /node_modules/,
-    loaders: ['react-hot', 'babel'],
```
bartaz: Moved to .babelrc as `react-hot-loader/babel` plugin.

src/common/root.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
```
bartaz: Had to move it out of `index.js` to be able to properly hot-reload it.

src/common/index.js
```diff
@@ -3,25 +3,28 @@ import './style/normalize.css';
...
-const component = (
-  <Router routes={routes} history={history} />
-);
+import Root from './root';
 
 render(
-  <Provider store={store}>
-    { component }
-  </Provider>,
+  <AppContainer>
```
bartaz: > <AppContainer/> is a component that handles module reloading, as well as error handling. The root component of your app should be nested in AppContainer as a child. When in production, AppContainer is automatically disabled, and simply returns its children.

from: https://github.com/gaearon/react-hot-loader/blob/next-docs/docs/README.md#migration-to-30

package.json
```diff
@@ -51,6 +51,7 @@
     "react-addons-test-utils": "^15.3.2",
     "react-dom": "^15.3.2",
     "react-helmet": "^3.1.0",
+    "react-hot-loader": "^3.0.0-beta.6",
```
bartaz: Because of `<AppContainer>`  that wraps whole application we need `react-hot-loader` in production dependencies.

> When in production, AppContainer is automatically disabled, and simply returns its children.

But it's still needed as dependency because of import.

src/common/index.js
```diff
@@ -3,25 +3,28 @@ import './style/normalize.css';
...
+    <Root />
+  </AppContainer>,
   document.getElementById('content')
 );
+
+if (module.hot) {
```
bartaz: As in: https://github.com/gaearon/react-hot-loader/blob/next-docs/docs/README.md#migration-to-30

bartaz:
Know issues with React Router:

> Currently, only synchronous routes are hot reloaded, and you will see a warning from <Router> on every reload. You can ignore this warning. For details, see: https://github.com/reactjs/react-router/issues/2182

from: https://github.com/gaearon/react-hot-boilerplate/blob/next/src/App.js#L5-L10

PR #222, opened 2016-11-22
Fsevents in shrinkwrap again

earnubs:
fsevents ended up in shrinkwrap again, breaking build in ubuntu; remove it, again.

fsevents was introduced here: https://github.com/canonical-ols/javan-rhino/pull/191/files#diff-36018a6a9e1de1798516799b84841318R952

... because the author had npm 3.10.3 installed, not 3.10.7


earnubs:
@bartaz sorry, forgot wip label, still investigating ... looking into whole prod vrs dev build too

vilagithub:
Looks like *avoiding* injecting fsevents is non-trivial (understatement).

But as long as we have a test to alert us when it happens, at least we should make progress on understanding when and why it's injected ?

So the test is definitely a step in the right direction and I'm +1 for landing this.

PR #223, opened 2016-11-28
Stripe failure handle with user message

earnubs:
handle failure when submitting a card to stripe  with a notification in the ui

### DIFF
--- a/src/common/actions/stripe.js
+++ b/src/common/actions/stripe.js
@@ -1,5 +1,6 @@
 import conf from '../config';
 import { postStripeToken } from './customer';
+import { showNotification } from './notifications';
 
 export const CREATE_STRIPE_TOKEN = 'CREATE_STRIPE_TOKEN';
 export const SAVE_VALIDATED_CARD_DATA = 'SAVE_VALIDATED_CARD_DATA';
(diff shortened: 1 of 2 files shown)


src/common/actions/stripe.js
```diff
@@ -28,9 +29,17 @@ export function createStripeTokenSuccess(token) {
 }
 
 export function createStripeTokenFailure(error) {
-  return {
-    type: CREATE_STRIPE_TOKEN_FAILURE,
-    error
+  return (dispatch) => {
+    dispatch(showNotification({
+      message: error,
```
bartaz: Shouldn't this be `message: error.message`?
I guess error in the response is not just string, but object with type, message, etc?

https://stripe.com/docs/api#errors

test/unit/src/common/actions/t_stripe.js
```diff
@@ -74,14 +76,29 @@ describe('stripe actions', () => {
...
+        }
+      }, {
         type: ActionTypes.CREATE_STRIPE_TOKEN_FAILURE,
         error
-      };
+      }];
+      const store = mockStore();
 
-      expect(createStripeTokenFailure(error)).toEqual(expectedAction);
+      store.dispatch(createStripeTokenFailure(error));
```
katiefenn: Can everything but the expectation be put in a `beforeEach` block? It makes reading code easier when code is folded 😊

Issue #217, opened 2016-11-21
Server requires dev dependencies when running on prod


bartaz:
https://github.com/canonical-ols/javan-rhino/blob/master/src/server/index.js#L3-L8

Our server requires couple of dev dependencies even when running on prod.
This means it will not run when dev dependencies are not installed.

If these are not needed in production environment, we should probably wrap them in NODE_ENV !== 'production' in this file.
If they are needed in prod (for example to build in production) we need to move them (and many more build dependencies out of devDependencies).

earnubs:
lines 6-8 will be needed in prod, but perhaps they could be written better to take advantage of dead code elimination? ie. If NODE_ENV is production, we don't need to require conf here (the path is just /static/icons)?

bartaz:
Another suspect `chokidar`.
We shouldn't be moving it to prod dependencies as it installs fsevents that may cause us trouble again when installing on Ubuntu.

https://github.com/canonical-ols/javan-rhino/blob/master/src/server/server.js#L7

katiefenn:
These changes work for me with `npm install && npm run build && npm install --production`:

https://github.com/canonical-ols/javan-rhino/compare/build-fix?expand=1

bartaz:
Fixed in #226

PR #225, opened 2016-11-30
ship only the production build of node_modules

earnubs:
running `npm i --production` after `npm i` removes devDeps from node_modules, and reduces the churn in staging diffs, as well as reducing the deployed code size


bartaz:
@earnubs This won't work because of #217

Currently we require some dev dependencies in server, so it wont run without them!

PR #226, opened 2016-12-01
separate dev and production servers

earnubs:
To keep development dependencies out of production I am proposing a new devel only server, I've also included a few tidy ups in the are of prod vrs dev.


src/server/index.js
```diff
@@ -1,14 +1,8 @@
-require('babel-register');
```
earnubs: https://github.com/babel/example-node-server/issues/13

package.json
```diff
@@ -37,11 +37,13 @@
     "classnames": "^2.2.5",
     "connect-memcached": "^0.2.0",
     "css-loader": "^0.26.0",
+    "css-modules-require-hook": "^4.0.5",
```
earnubs: I would be interested at some point in looking into a babel plugin for these require hooks, so that it was handled at transpilation time.

npm-shrinkwrap.json
```diff
@@ -3,1094 +3,1867 @@
   "version": "1.0.0",
   "dependencies": {
     "accepts": {
-      "version": "1.3.3"
+      "version": "1.3.3",
+      "from": "accepts@1.3.3",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz"
```
bartaz: `npm-shrinkwrap.json` seems to be not shonkrapped (contains `from` and `resolved` fields.

Yet still it seems to be passing the npm-shrinkwrap tests that check for `resolved`?

PR #227, opened 2016-12-06
Google analytics

earnubs:
add ga.js to capture page views and associated stats

### DIFF
--- a/src/common/root.js
+++ b/src/common/root.js
@@ -13,6 +13,18 @@ const component = (
   <Router routes={routes} history={history} />
 );
 
+history.listen((location) => {
+  if ('undefined' !== typeof window) {
+    const ga = window.ga;
+    if (ga) {
+      ga('send', {
+        hitType: 'pageview',
+        page: location.pathname
+      });
+    }
+  }
+});
+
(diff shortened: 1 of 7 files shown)


src/common/routes.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
-import { Route, Redirect } from 'react-router';
-import App from './containers/app.js';
 import AddCard from './containers/add-card.js';
+import App from './containers/app.js';
```
earnubs: Pardon the churn, I sorted theses alphabetically because I was addind a new dep (react-ga) and then abandoned it (react-ga not happy in universal app).

src/common/root.js
```diff
@@ -13,6 +13,18 @@ const component = (
   <Router routes={routes} history={history} />
 );
 
+history.listen((location) => {
+  if ('undefined' !== typeof window) {
```
katiefenn: Why the Yoda condition?

earnubs: Like them, I do.

... but I think this conditional can be removed altogether since browserHistory depends on the History API, so if this event is fired we have a `window` object.

PR #230, opened 2016-12-12
add phantomjs

earnubs:
npm won't pull phantomjs from the network if it's installed already, so this pr (might) avoid proxy issues in jenkins builds


vilagithub:
All good !

PR #232, opened 2016-12-12
remove phantomjs, use chrome

earnubs:
Until we switch to browserstack tests, use locally installed chrome for testing, because phantomjs install on ubuntu is a little more complex than I care for right now (and will go away RSN with browserstack integration).


dependencies-devel.txt
```diff
@@ -1 +0,0 @@
-phantomjs
```
vilagithub: You may as well remove the file and its reference in ols-vms.conf

earnubs: good spot, done!

PR #236, opened 2017-01-17
use webdriver proxy

earnubs:
use webdriver proxy; proxy is null in the constructor, so if the env var isn't set it's NOOP

### DIFF
--- a/test/acceptance/www/payment/edit/t_edit.js
+++ b/test/acceptance/www/payment/edit/t_edit.js
@@ -65,6 +65,7 @@ for (let browser of browsers) {
           ...capabilities,
           ...browser
         })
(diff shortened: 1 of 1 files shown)


test/acceptance/www/payment/edit/t_edit.js
```diff
@@ -60,11 +60,12 @@ for (let browser of browsers) {
       }
 
       driver = new Builder()
-        .usingServer('http://hub.browserstack.com/wd/hub')
+        .usingServer('https://hub.browserstack.com/wd/hub')
         .withCapabilities({
           ...capabilities,
           ...browser
         })
+        .usingWebDriverProxy(process.env.HTTPS_PROXY)
```
vilagithub: shoot in the dark: shouldn't that come before usingServer() ?

PR #250, opened 2017-01-27
no need for statsd string here

earnubs:
* in graphite stats are appearing in 'statsd.statsd' so we can drop one of those :)
* set the scope to be 'ols'.
* create client instance once.


earnubs:
@Roadmaster no entirely sure, when this current crunch is over I'd like to take a closer look, or maybe switch to `prom-client`

PR #254, opened 2017-02-01
Add JUJU_UNIT and SERVICE_ENVIRONMENT variables

Roadmaster:
These variables will be set in the environment available to the service, via the systemd file's Environment directives.

JUJU_UNIT will have any slashes replaced by hyphens (so javan-rhino/0 becomes javan-rhino-0 as is conventional for e.g. usage in statsd identifiers).


Roadmaster:
Self-comment> please don't merge this since it looks like the approach is not great. I'll submit a new PR tomorrowish.

PR #257, opened 2017-02-09
Add extra environment variables (incl. JUJU_UNIT and SERVICE_ENVIRONMENT)

Roadmaster:
This renders the common extra variables (including JUJU_UNIT and SERVICE_ENVIRONMENT among others) defined by the OLS layer (and possibly others) into a file in the service's etc/ directory. This file can trivially be used via EnvironmentFile in the systemd unit (also done here) or sourced manually if the service needs to be diagnosed or otherwise run manually. This approach should be closer to recommended best practice on the OLS layer family and allows for easily moving the existing Environment= variables to the env file. I didn't do that but can add that in a subsequent commit if reviewers think it's worth doing.


bloodearnest:
Looks good, main thing was to reuse env_vars() some how, this works

PR #259, opened 2017-02-13
add terms of service copy

earnubs:
Source available here https://docs.google.com/document/d/1bNNqZJqjK-rPEtky2XAMyh0j0lBnlLG27ws50hzroU4/edit#

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-dist: trusty
-sudo: false
 language: node_js
 node_js:
   - "6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ node_js:
   - "6"
   - "node"
 after_success: npm run coverage
-before_install: npm i -g npm@3.10.7;
+#before_install: npm i -g npm@3.10.7;
--- a/src/common/containers/terms.js
+++ b/src/common/containers/terms.js
@@ -61,7 +61,7 @@ export default class Terms extends Component {
         </p>
 
         <p>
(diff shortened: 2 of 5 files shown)


src/common/containers/terms.js
```diff
@@ -3,66 +3,426 @@ import React, { Component } from 'react';
...
+        <h3>2. Services and Software</h3>
+
+        <p>
+          <strong>The Service enables you to download and install snaps
+            and software (“Software”) on Snappy Ubuntu Core and any
+            compatible supported systems.</strong>
+        </p>
+
+        <p>
+          Details of the Services can be found here: <a href={ SNAPCRAFT_URL }>http://snapcraft.io/</a>
```
Roadmaster: If the snapcraft URL is configurable, would it make sense to <a href={ SNAPCRAFT_URL }>{SNAPCRAFT_URL}</a>? 
I understtand the legal text shouldn't be modified, but the possibility for mismatch and the wasted opportunity for DRYness hurt my brain :(

Roadmaster:
Will this supersede https://github.com/canonical-ols/my.ubuntu.com/pull/143 ?

.travis.yml
```diff
@@ -1,5 +1,3 @@
-dist: trusty
-sudo: false
```
earnubs: Tests keep timing out on travis, which incidently has been up to 3x slower since I switched to containers, so I tried this to see if it fixed the test (it didn't) but will keep it because the switch to containers didn't help test times,

PR #261, opened 2017-02-17
Reload systemd units

Roadmaster:
When the config changes, it's not really applied (even if systemctl restarting the service) unless systemd is told to reload the units. This fixes that by enabling the service (just in case) and reloading units.


charm/reactive/javan_rhino.py
```diff
@@ -44,6 +53,8 @@ def configure(cache):
                 'env_file': env_file,
                 'environment': environment,
             })
+        check_call(['systemctl', 'enable', basename(SYSTEMD_CONFIG)])
```
Roadmaster: These two check_call lines are the meat of it. The rest are pep8 fixes and imports.

PR #278, opened 2018-10-16
Remove some unused dependencies

cjwatson:


Roadmaster:
LGTM. The changes to shrinkwrap scare me a bit but I did check it and it's consistent with the dependencies you removed so it should be fine.

PR #280, opened 2018-10-17
Switch to babel-preset-env

cjwatson:
babel-preset-es2015 now emits a warning telling us to use
babel-preset-env instead, so follow its advice.


cjwatson:
It shouldn't be like the pip/wheel stuff, no.  The babel machinery is all in JS, and AFAICT works fine with the version of node we're using.  (Also, BSI has been on babel-preset-env for a while already.)

